PR #6, opened 2016-03-04
Fix compilation for IDLs that use relative paths in includes

### LINKED ISSUE #5: Compilation breaks for IDLs with use includes with relative paths
# Observed Behaviour

The thrifty compiler fails when a Thrift IDL file declares includes using relative paths. So for example: there is a file `Foo.thrift` in the `foo` folder that depends on another file `Bar.thrift` in the `bar` folder and it declares the dependency like this: `include ../bar/Bar.thrift`. When running `java -jar thrifty-compiler-0.2.0.jar ./foo/Foo.thrift ./bar/Bar.thrift` the compiler will fail with the following exception:
```
Exception in thread "main" java.lang.AssertionError: All includes should have been resolved by now: /some/absolute/path/foo/../bar/Bar.thrift
	at com.microsoft.thrifty.schema.Loader.getAndCheck(Loader.java:274)
	at com.microsoft.thrifty.schema.Loader.resolveIncludedProgram(Loader.java:232)
	at com.microsoft.thrifty.schema.Program.loadIncludedPrograms(Program.java:232)
	at com.microsoft.thrifty.schema.Loader.loadFromDisk(Loader.java:145)
	at com.microsoft.thrifty.schema.Loader.load(Loader.java:107)
	at com.microsoft.thrifty.compiler.ThriftyCompiler.compile(ThriftyCompiler.java:178)
	at com.microsoft.thrifty.compiler.ThriftyCompiler.main(ThriftyCompiler.java:93)
```

# Cause

The `loadFromDisk` method in the `Loader` class stores `Program`s using their canonical paths but `getAndCheck` is invoked using the absolute path built by the `findFirstExisting` method, which creates a file using parent and file paths when resolving paths according to the current location or include paths rules.

# Fix

Invoking the `getAndCheck` using canonical paths instead of absolute ones.

stefanodacchille:
Fixes #5


thrifty-integration-tests/bar/Bar.thrift
```diff
@@ -0,0 +1,24 @@
+/*
```
benjamin-bader: Let's stick with the Microsoft copyright header; this Apache Foundation header is on a few specific .thrift files that we have taken from the Apache implementation, but new files should not be thusly attributed.

thrifty-integration-tests/build.gradle
```diff
@@ -56,7 +56,17 @@ task compileTestThrift(type: Exec) {
     args('-jar', jarTask.archivePath.absolutePath, "--out=$projectDir/build/generated-src/thrifty", "$projectDir/ClientThriftTest.thrift")
 }
 
-tasks['compileTestJava'].dependsOn compileTestThrift
+
+task compileTestIncludeThrift(type: Exec) {
```
benjamin-bader: I think it might be easier if, instead of using a custom Gradle task, this test were added to `LoaderTest.java` in `thrifty-schema`; adding Gradle task dependencies makes me nervous as they are picky and easy to break.

thrifty-integration-tests/bar/Bar.thrift
benjamin-bader: Actually, per my other comment, this and `Foo.thrift` should go away, and the test moved to a junit case in `LoaderTest.java`

benjamin-bader:
Thanks for contributing!  This is a very straightforward change; the only change I'd request is moving the test from a custom Gradle task to a junit test case in `thrifty-schema/src/test/.../LoaderTest.java`, which is set up for just this sort of thing.

Once that happens and you do the CLA thing, I'll be more than happy to land this.

stefanodacchille:
No worries! I made the changes you proposed. I didn't see those unit tests for the `Loader` class, I'm completely new to the codebase. I agree it's much better than relying on gradle tasks.

I've got to clarify to whole CLA thing with my current employer, shouldn't be an issue though.

stefanodacchille:
Not sure why the integration tests are failing. I performed a clean build locally and everything works for me.

thrifty-schema/src/test/java/com/microsoft/thrifty/schema/LoaderTest.java
```diff
@@ -358,6 +359,35 @@ public void canLoadAndLinkOfficialTestThrift() throws Exception {
...
+        File f1 = tempDir.newFile("a.thrift");
+        File f2 = tempDir.newFile(Joiner.on(File.separatorChar).join("foo", "b.thrift"));
+
+        String a = "namespace java com.microsoft.thrifty.test.includesWithRelativePaths\n" +
+                "\n" +
+                "enum A {\n" +
+                "  ONE, TWO, THREE\n" +
+                "}";
+
+        String b = "include '" + f1.getCanonicalPath() + "'\n" +
```
benjamin-bader: I don't think this tests what you want it to; did you mean to say `include '../a.thrift'`?

benjamin-bader:
Awesome!  One more comment, then it's just up to the CLA process.  There aren't any surprises in there, so hopefully your employer is understanding.

Re: the test failure, the integration tests are just janky right now and sometimes time out on Travis.  I'm not sure what the actual issue is, but the tests involve local TCP communication so who knows which part of the CI system might be slowing that down.  I'll open an issue to investigate.

thrifty-schema/src/test/java/com/microsoft/thrifty/schema/LoaderTest.java
stefanodacchille: Yes that's what I should have done, thanks. Just amended and pushed the changes.

benjamin-bader:
Thanks!

stefanodacchille:
Glad I was able to help!

PR #33, opened 2016-06-24
Thrifty compiler bugs

chaviw:
Fix a few thrifty compiler bugs:
1. Allow different ThriftTypes with the same name to work. This is valid in Java so should be allowed.
2. Fix issue with TypeDef enums where the compiler failed to recognize the value as a valid enum

### DIFF
--- a/thrifty-java-codegen/src/test/java/com/microsoft/thrifty/gen/ThriftyCodeGeneratorTest.java
+++ b/thrifty-java-codegen/src/test/java/com/microsoft/thrifty/gen/ThriftyCodeGeneratorTest.java
@@ -90,6 +90,61 @@ public class ThriftyCodeGeneratorTest {
                 .compilesWithoutError();
     }
 
+    /**
+     * Tests for duplicate names of different ThriftTypes.
+     * Example: an enum and a constant can have the same name.
+     * @throws Exception
+     */
+    @Test
+    public void testDuplicateNamesDiffTypes() throws Exception {
+        String thrift = Joiner.on('\n').join(
+                "namespace java test",
+                "",
+                "const string Foo = \"foo\"",
+                "enum Foo {",
+                "  Bar",
+                "  Bar2",
+                "}");
+
+        Schema schema = parse("foobar.thrift", thrift);
+
+        ThriftyCodeGenerator gen = new ThriftyCodeGenerator(schema);
+        ImmutableList<JavaFile> javaFiles = gen.generateTypes();
+
+        assertThat(javaFiles).hasSize(2);
+
+        assertAbout(javaSource())
+                .that(javaFiles.get(0).toJavaFileObject())
+                .compilesWithoutError();
+    }
+
+    /**
+     * Tests for case where a typedef enum is set to a constant.
+     * @throws Exception
+     */
+    @Test
+    public void testTypeDefEnumAsConstant() throws Exception {
+        String thrift = Joiner.on('\n').join(
+                "namespace java test",
+                "",
+                "enum Foo {",
+                "  Bar",
+                "  Bar2",
+                "} typedef Foo FooType",
+                "const FooType foo = Foo.Bar");
+
+        Schema schema = parse("typedef_enum.thrift", thrift);
+
+        ThriftyCodeGenerator gen = new ThriftyCodeGenerator(schema);
+        ImmutableList<JavaFile> javaFiles = gen.generateTypes();
+
+        assertThat(javaFiles).hasSize(2);
+
+        assertAbout(javaSource())
+                .that(javaFiles.get(0).toJavaFileObject())
+                .compilesWithoutError();
+    }
+
     private Schema parse(String filename, String text) throws Exception {
         File file = tmp.newFile(filename);
         PrintWriter writer = new PrintWriter(file, "UTF-8");
--- a/thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Constant.java
+++ b/thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Constant.java
@@ -258,6 +258,12 @@ public class Constant extends Named {
                 } else {
                     throw new IllegalStateException("bad enum literal: " + value.value());
                 }
+            } else if (named instanceof Typedef) {
+                Typedef td = (Typedef) named;
+                if (td.oldType().isEnum()) {
+                    return;
+                }
+                throw new IllegalStateException("Invalid type");
             } else if (named instanceof Constant) {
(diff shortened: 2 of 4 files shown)


thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Constant.java
```diff
@@ -258,6 +258,12 @@ public void validate(Linker linker, ThriftType expected, ConstValueElement value
                 } else {
                     throw new IllegalStateException("bad enum literal: " + value.value());
                 }
+            } else if (named instanceof Typedef) {
```
benjamin-bader: If we do the typedef check this way, it looks like we'll miss validating the value of the enum.  How will we validate that the aliased type is the correct enum type?

I think a better approach would be to ensure that we don't look up a typedef in the first place.

thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Program.java
```diff
@@ -240,14 +242,52 @@ void loadIncludedPrograms(Loader loader, Set<Program> visited) throws IOExceptio
...
+        namedList.clear();
+        namedList.addAll(constants);
+        checkList(namedList, symbolMap);
+
+        this.symbols = ImmutableMap.copyOf(symbolMap);
+    }
+
+    private void checkList(Iterable<Named> list, LinkedHashMap<String, Named> symbolMap) {
+        LinkedHashMap<String, Named> localMap = new LinkedHashMap<>();
+        for (Named enumName : list) {
```
benjamin-bader: nit: `enumName` -> `name`

thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Program.java
```diff
@@ -240,14 +242,52 @@ void loadIncludedPrograms(Loader loader, Set<Program> visited) throws IOExceptio
...
 
         LinkedHashMap<String, Named> symbolMap = new LinkedHashMap<>();
-        for (Named named : names()) {
-            Named oldValue = symbolMap.put(named.name(), named);
+
+        // Some type-resolution subtlety eludes me.  I'd have thought that
+        // Iterable<EnumType> is castable to Iterable<Named> (inheritance),
+        // but the IDE claims otherwise.  So, instead create a temp list where each list's
+        // data will be copied into
+        List<Named> namedList = new ArrayList<>();
```
benjamin-bader: I agree that we need to separate duplicate-symbol-detection of constants from that of types.  That said, this seems to be too loose - we will only detect name clashes within one particular subtype of `Named`; a `struct` and a `union` with the same name is definitely not legal, but will pass validation this way.

We ought to just do two passes on `names()`, one where we check constants and another where we check non-constants.

On a deeper level this change raises a troubling issues - the way we store and query symbols is now shown to be broken.  A name could refer to either a constant or a type, and we currently don't allow that.  This requires some thought...

benjamin-bader: Also, why bother with another list?  I think that we can avoid it (and the covariance issue in the comment) by making the `checkList` method appropriately generic:

```
private void checkList(Iterable<? extends Named> names, ...)
```

benjamin-bader:
Hi, thanks for your contribution!  You raise a few good bugs here - constant names indeed should not clash with type names.  The error reporting could definitely use some love as well, as you noticed.  I'm also very happy to see tests added too!

As noted in one of the comments, this raises an issue that is now bothering me - the way we store and query symbols during linking and codegen now seems inadequate.  I'll have to investigate.

thrifty-java-codegen/src/test/java/com/microsoft/thrifty/gen/ThriftyCodeGeneratorTest.java
```diff
@@ -90,6 +90,61 @@ public void enumGeneration() throws Exception {
...
+    @Test
+    public void testTypeDefEnumAsConstant() throws Exception {
+        String thrift = Joiner.on('\n').join(
+                "namespace java test",
+                "",
+                "enum Foo {",
+                "  Bar",
+                "  Bar2",
+                "} typedef Foo FooType",
+                "const FooType foo = Foo.Bar");
```
benjamin-bader: We should add a test that assigns an enum value of the wrong type to the typedef, i.e.

```
enum Foo {
  ...
}

enum Bar {
  ...
}

typedef Foo FooType

const FooType Bar.SomeValue
```

thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Constant.java
chaviw: What do you mean "don't look up a typedef"? Do you mean don't classify them as enums? What validation would typdef enums have at all then?

thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Program.java
chaviw: You're correct with the second point. I should have done that. I can make that change

I'm gonna think about the first issue you mentioned. Seems like the parser might have to get a little more complicated.

chaviw:
Our thrift files actually had these cases so it broke when trying to compile the thrift files. I hope we can find a solution to the constant/type name clashing because our team was hoping to switch from thrift to thrifty.

benjamin-bader:
Leaving the const/type confusion aside, we can patch the bugs you've raised and thrifty will *most likely* just work for you - but it will still be flawed.  I'm open to letting it be flawed in this way for a little while because it doesn't seem that this will bite very many users at the moment.

thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Constant.java
benjamin-bader: Usually we use `type.getTrueType()` to resolve typedefs; we just need to do that to the type being validated.

I think the bug is actually on line 79.  When we look up a validator based on its type, the first thing that happens is that we resolve the typedef.  When we then call `validate`, we pass the original, unresolved, type.  So line 79 ought to say:

```java
Validators.forType(expectedType).validate(linker, expectedType.getTrueType(), value);
```

or, split it up:

```java
expectedType = expectedType.getTrueType();
Validators.forType(expectedType).validate(linker, expectedType, value);
```

This way, when we later look up a `Named` instance for the given type, we will get back the `EnumType` and not the `Typedef`.

benjamin-bader:
All of the bugs mentioned here have since been fixed or obviated, so I'm closing this PR as obsolete.  Thank you for the effort!

Issue #39, opened 2016-07-15
Proposal: Make generated structs implement an interface


seanabraham:
My use case involves serializing any struct without knowing the specific class. In order to do this, they need to implement a common interface. In this case, I imagine the interface will be something like the following:

```java
public interface Struct<T, B extends StructBuilder<T>> {
    Adapter<T, B> getAdapter();
}
```

Then a given generated would look like the following:

```java
public final class Foo implements Struct<Foo, Foo.Builder> {
  public static final Adapter<Foo, Builder> ADAPTER = new FooAdapter();

  ...
  
  @Override
  public Adapter<Foo, Builder> getAdapter() {
    return ADAPTER;
  }

  ...
}
```

and you'd be able to serialize a generic struct via something like:

```java
<T extends Struct<T, B>, B extends StructBuilder<T>> void serialize(T struct) {
  Buffer buffer = new Buffer();
  Transport transport = new BufferTransport(buffer);
  Protocol protocol = new BinaryProtocol(transport);

  struct.getAdapter().write(protocol, struct)
}
```

Since generated structs don't currently extend from or implement any interfaces, I don't imagine this would cause any breakage.

I'm happy to submit a PR that accomplishes this, I believe it would be pretty straight forward.

benjamin-bader:
Serializing generic thrifts would be a very useful feature; we didn't do it initially to save the extra method per class, but I'm not at all opposed in principle.

I'd be happy to take a PR.

seanabraham:
Actually on second thought, a Struct itself wouldn't need to return the whole Adapter because if you already have an instance of the struct you wouldn't be doing any reading. It would just need to write out to a Protocol so this interface can be simplified a bit. I'll put up a PR shortly with that in mind! Let me know what you think.

PR #40, opened 2016-07-15
Add the Struct Interface and make generated Thrift structs implement it

seanabraham:
Per https://github.com/Microsoft/thrifty/issues/39 this adds the `Struct` interface which has a `write` method and makes all generated Thrift structs implement it. This will add one method per generated struct but with the advantage that a generic `Struct` can be written out to a protocol.

For example, this is now possible:

```java
Buffer buffer = new Buffer();
Transport transport = new BufferTransport(buffer);
Protocol protocol = new BinaryProtocol(transport):

Struct fooStruct = Foo.Builder().build();
Struct barStruct = Bar.Builder().build();

fooStruct.write(protocol);
barStruct.write(protocol);
```

Paired with @felipecsl


benjamin-bader:
Hi, thanks for the speedy PR and CLA!

It looks like this failed checkstyle; please run `./gradlew check` and fix whatever it is complaining about, usually it's very minor.

This is quite useful as-is; now that I see this, however, I'm wondering if it might be better to have `Struct` as a common base class rather than an interface.  It would then be trivial to say:
```java
protected abstract Adapter<T, B> adapter();
public void serialize(Protocol protocol) {
  adapter().write(protocol, (T) this);
}
```

Not a huge difference yet, but it would lay the groundwork for adding common functions to generated classes without codegen.

What do you think?

felipecsl:
+1 for that suggestioN!

seanabraham:
Good idea, I'll fix the checkstyle real quick (missing period in javadoc) and then modify the change to extend from `Struct` instead.

seanabraham:
So, one issue with extending from a `Struct` abstract class as opposed to implementing a `Struct` interface is the existing behavior in the `buildStruct` method. It looks like if the `StructType` is `StructElement.Type.EXCEPTION` the generated class is meant to be a subclass of `Exception`:

```
    if (type.isException()) {
      structBuilder.superclass(Exception.class);
    }
```

benjamin-bader:
Of course - how could I forget!  That certainly complicates things.

It's really a shame that, even with the new Java 8 features coming in Android N, default methods on interfaces will not be supported below API 24; that would give us the best of all worlds.  I suppose we can proceed with an interface.

Thanks for your contribution!

PR #48, opened 2016-07-30
Preserve annotations on type references

benjamin-bader:
This allows arbitrary type-references (that is, type names that are part of definitions of other types) to keep retain thrift annotations after linking.

Prior to this change, we canonicalized ThriftTypes into single instances, and a byproduct of that process was the elimination of these annotations.  Now we can keep them and use them in codegen.

This is a better solution to #43.

### DIFF
--- a/thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Linker.java
+++ b/thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Linker.java
@@ -20,6 +20,7 @@
  */
 package com.microsoft.thrifty.schema;
 
+import com.microsoft.thrifty.schema.parser.AnnotationElement;
 import com.microsoft.thrifty.schema.parser.ListTypeElement;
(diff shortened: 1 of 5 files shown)


thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Linker.java
```diff
@@ -292,42 +293,48 @@ private void register(Named type) {
...
+        Map<String, String> annotations = annotationElement != null
+                ? annotationElement.values()
+                : Collections.<String, String>emptyMap();
+
         ThriftType tt = typesByName.get(type.name());
         if (tt != null) {
-            return tt;
+            // If we are resolving e.g. the type of a field element, the type
+            // may carry annotations that are not part of the canonical type.
+            if (annotations.isEmpty()) {
```
benjamin-bader: should be *not* `annotations.isEmpty()`.

Add some unit tests to verify, Travis shouldn't have let this through.

PR #50, opened 2016-08-02
Add validation for Services at link-time.

### LINKED ISSUE #45: Validate service definitions at link-time
[Linker has had a stub](https://github.com/Microsoft/thrifty/blob/master/thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Linker.java#L286) for this forever.  Currently, Thrifty will happily accept all manner of nonsensical service definitions, like `oneway` functions that return data.  It will be relatively simple to add validation - just a matter of adding validation on `Service` and `ServiceMethod`.  Validation should verify that:

For services:
* A service either has no base type, or if it does, that base type is also a service

For service methods:
* A service method has a non-`void` return type and is not marked one-way *or* it is marked one-way and has a `void` return type
* All `exceptionTypes` are, in fact, exceptions; this can be verified by using `Linker#lookupSymbol(ThriftType)` and checking the resulting `StructType`.

Most of the other `Named` subtypes (e.g. `StructType`, `Constant`) have `validate(Linker)` methods that can serve as an example.

benjamin-bader:
Fixes #45.

### DIFF
--- a/thrifty-schema/src/main/java/com/microsoft/thrifty/schema/ServiceMethod.java
+++ b/thrifty-schema/src/main/java/com/microsoft/thrifty/schema/ServiceMethod.java
@@ -60,6 +60,10 @@ public final class ServiceMethod {
         this.annotations = annotationBuilder.build();
     }
 
+    public Location location() {
+        return element.location();
+    }
+
     public String documentation() {
         return element.documentation();
     }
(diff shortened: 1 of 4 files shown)


thrifty-schema/src/main/java/com/microsoft/thrifty/schema/ServiceMethod.java
```diff
@@ -107,4 +111,36 @@ void link(Linker linker) {
...
+
+        for (Field field : exceptionTypes) {
+            ThriftType type = field.type();
+
+            if (type.isBuiltin()) {
+                linker.addError(field.location(), "Only exception types can be thrown");
+            } else {
+                Named named = linker.lookupSymbol(type);
+                if (named == null) {
+                    throw new AssertionError("wtf");
```
benjamin-bader: Add a more appropriate error here

PR #61, opened 2016-09-07
Add Builders to Schema Components 

naturalwarren:
Adds `toBuilder()` API to schema components . This allows for creating custom instances of `Schema` components while respecting the original `Schema`'s immutability.

Overview of Changes:

`toBuilder()`
- Takes an instance of a class, returns a builder populated with the class’s fields

Builder Classes
- Added to each of the classes that are a field in `Schema`
- Builder is not publicly constructible, can only create a new one from an existing component instance

Putting this up to get feedback now. Happy to iterate on documentation and API surface area as needed. 

Big thanks to @hzsweers for thinking this up and helping implement it.

### DIFF
--- a/thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Constant.java
+++ b/thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Constant.java
@@ -90,6 +90,24 @@ public class Constant extends Named {
         Validators.forType(trueExpectedType).validate(linker, trueExpectedType, value);
     }
 
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) { return true; }
+        if (o == null || getClass() != o.getClass()) { return false; }
+
+        Constant constant = (Constant) o;
+
+        if (element != null ? !element.equals(constant.element) : constant.element != null) { return false; }
+        return type != null ? type.equals(constant.type) : constant.type == null;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = element != null ? element.hashCode() : 0;
+        result = 31 * result + (type != null ? type.hashCode() : 0);
+        return result;
+    }
+
     interface ConstValueValidator {
         void validate(Linker linker, ThriftType expected, ConstValueElement value);
     }
@@ -108,6 +126,9 @@ public class Constant extends Named {
         }
 
         public Builder element(ConstElement element) {
+            if (element == null) {
+                throw new NullPointerException("element may not be null.");
+            }
             this.element = element;
             return this;
         }
--- a/thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Constant.java
+++ b/thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Constant.java
@@ -38,7 +38,7 @@ public class Constant extends Named {
         this.namespaces = namespaces;
     }
 
-    public Constant(Builder builder) {
+    private Constant(Builder builder) {
         super(builder.element.name(), builder.namespaces);
         this.element = builder.element;
         this.namespaces = builder.namespaces;
@@ -97,12 +97,12 @@ public class Constant extends Named {
         void validate(Linker linker, ThriftType expected, ConstValueElement value);
     }
 
-    private static final class Builder {
+    public static final class Builder {
         private ConstElement element;
         private Map<NamespaceScope, String> namespaces;
         private ThriftType type;
 
-        public Builder(ConstElement element,
+        Builder(ConstElement element,
                        Map<NamespaceScope, String> namespaces,
                        ThriftType type) {
             this.element = element;
--- a/thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Constant.java
+++ b/thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Constant.java
@@ -29,11 +29,20 @@ import java.util.Map;
 
 public class Constant extends Named {
     private final ConstElement element;
+    private final Map<NamespaceScope, String> namespaces;
     private ThriftType type;
 
     Constant(ConstElement element, Map<NamespaceScope, String> namespaces) {
         super(element.name(), namespaces);
         this.element = element;
+        this.namespaces = namespaces;
+    }
+
+    public Constant(Builder builder) {
+        super(builder.element.name(), builder.namespaces);
+        this.element = builder.element;
+        this.namespaces = builder.namespaces;
+        this.type = builder.type;
     }
 
     @Override
@@ -52,6 +61,10 @@ public class Constant extends Named {
 
     public ConstValueElement value() {
         return element.value();
+    }
+
+    public Builder toBuilder(Constant constant) {
+        return new Builder(constant.element, constant.namespaces(), constant.type);
     }
 
     void link(Linker linker) {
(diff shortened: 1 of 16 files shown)


benjamin-bader:
Hi, thanks for the contribution!  I'm excited to get this landed, and to see what we can build on top of it!

Per @hzsweers it sounds like this is still a work in progress, so I won't get too far down into the weeds at this point.  As a general point of style though, I tend to avoid `set` prefixes on methods of builder classes; we set the precedent in some of the AutoValue builders in the parser.  I'd like to continue with that style here.

The only other general comment here is that it would be awesome to include some unit tests for the new builders, ensuring e.g. that `thing.toBuilder().build().equals(thing)`.

thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Constant.java
```diff
@@ -54,6 +63,10 @@ public ConstValueElement value() {
         return element.value();
     }
 
+    public Builder toBuilder(Constant constant) {
```
benjamin-bader: The other new builders all share this same trait, so I'll just comment once here.

`toBuilder` here is an instance method on `Constant`, but it's taking another `Constant` as an argument.  Any reason we couldn't eliminate the formal parameter and just use `this` instead?  For example, this would be simpler:

```java
public Builder toBuilder() {
  return new Builder(element, namespaces(), type);
}
```

thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Constant.java
```diff
@@ -29,11 +29,20 @@
 
 public class Constant extends Named {
     private final ConstElement element;
+    private final Map<NamespaceScope, String> namespaces;
```
benjamin-bader: It looks like this is duplicating the base class field `Named#namespaces` - what's the rationale for that?  Do we need to be able to assign to the field from here, or can we get by with accessing it from the `namespaces()` method?  If not, let's consider changing the original field from `private` to `protected`.

thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Constant.java
hzsweers: Agreed, `toBuilder()` should have no arguments

naturalwarren:
Hey Ben I'm excited for this to go in too! Right now I've got a lot on my plate and need to Revisit this on Monday. I hope this doesn't inconvenience you. I'd like to leave the pull request open and pick it up then. Let me know what you think. - Warren

benjamin-bader:
Cool, no rush.  Thanks again!

naturalwarren:
Hey Ben. Addressed everything inline except didn't add tests.

Are you happy with the surface area of this change (are more builders needed in other places)? If so I'll write some tests. If not what else would you like to see?

Zac mentioned you wanted to see a sample of what `toBuilder()` could be used for. I'll mock that up and paste something here.

naturalwarren:
Here's an example of how these API's would be used. This is a little arbitrary so let me know if you'd like something more concrete. 

```
package com.microsoft.thrifty.schema;

import com.google.common.collect.ImmutableList;
import com.google.common.collect.ImmutableMap;

import java.util.ArrayList;
import java.util.HashMap;
import java.util.List;
import java.util.Map;

import autovalue.shaded.com.google.common.common.collect.Maps;

/**
 * Examines annotations on an {@link ServiceMethod}.
 * Adds the annotation generated_class if the annotation http.status is present.
 */
public class AnnotationProcessor {

    private static final Map<String, StructType> exceptionByExceptionTypeName = new HashMap<>();

    /**
     * Adds generated_class annotation to a {@link ServiceMethod}'s annotations if http.status annotation is present.
     *
     * @param schema the schema to be processed.
     * @return a modified schema with added generated_class annotations.
     */
    public Schema processSchema(Schema schema) {
        Schema.Builder schemaBuilder = schema.toBuilder();

        // Populate map with whatever data is required to add generated_class exception.
        for (StructType type : schema.exceptions()) {
            exceptionByExceptionTypeName.put(type.type().name(), type);
        }

        List<Service> services = new ArrayList<>();
        List<ServiceMethod> serviceMethods = new ArrayList<>();
        HashMap<String, String> annotationMap = Maps.newHashMap();

        for (Service service : schema.services()) {
            Service.Builder serviceBuilder = service.toBuilder();
            for (ServiceMethod serviceMethod : service.methods()) {
                ServiceMethod.Builder serviceMethodBuilder = serviceMethod.toBuilder();
                
                annotationMap.putAll(serviceMethod.annotations());
                if (serviceMethod.annotations().containsKey("http.status")) {
                    annotationMap.put("generated_service_exception_name",
                            exceptionByExceptionTypeName.get(serviceMethod.name()).name());
                }
                serviceMethodBuilder.annotations(ImmutableMap.copyOf(annotationMap));
                serviceMethods.add(serviceMethodBuilder.build());
            }
            serviceBuilder.methods(ImmutableList.copyOf(serviceMethods));
            services.add(serviceBuilder.build());
        }
        
        return schemaBuilder.build();
    }
}
```

thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Schema.java
```diff
@@ -107,4 +117,73 @@ public EnumType findEnumByType(ThriftType type) {
...
+            this.structs = structs;
+            this.unions = unions;
+            this.exceptions = exceptions;
+            this.enums = enums;
+            this.constants = constants;
+            this.typedefs = typedefs;
+            this.services = services;
+        }
+
+        public Builder structs(ImmutableList<StructType> structs) {
```
benjamin-bader: Commenting once here for all of methods on this builder: we need to check the argument for null here, throwing `NullPointerException` if so.

thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Service.java
```diff
@@ -90,6 +98,57 @@ public ThriftType extendsService() {
...
+            this.type = type;
+            this.annotations = annotations;
+        }
+
+        public Builder element(ServiceElement element) {
+            this.element = element;
+            return this;
+        }
+
+        public Builder methods(ImmutableList<ServiceMethod> methods) {
```
benjamin-bader: Check for null here

thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Schema.java
naturalwarren: Done.

thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Service.java
naturalwarren: Done.

naturalwarren:
Addressed Ben's comments. Added tests.

naturalwarren:
Generated `equals()` and `hashCode()` methods to classes with builders so they can be properly tested.

thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Schema.java
naturalwarren: https://github.com/Microsoft/thrifty/pull/61/commits/969a978c2d961da58b9e5195ea7aef0544db0937

thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Service.java
naturalwarren: https://github.com/Microsoft/thrifty/pull/61/commits/d76400f739dcb228574b09021d7411a87eb9dad4

naturalwarren:
Good catch on the names. Fix in https://github.com/Microsoft/thrifty/pull/61/commits/0e96b5376624955e8268955dbeae1740f6b27084.

benjamin-bader:
Merged - thanks for all your good work!

PR #62, opened 2016-09-26
Name is state altering, remove it from hashcode.

naturalwarren:
Calling `name()` alters the `javaName` property. Remove this property from `hashcode()`.

Noticed this while bumping our model library onto the latest Thrifty. We're consuming Builders from #61 :)

cc @hzsweers


naturalwarren:
I'll remove `javaName` from `equals()` as well

benjamin-bader:
LGTM, thanks!

Issue #69, opened 2016-10-17
Unify Named with ThriftType


benjamin-bader:
As time passes, the distinction seems less and less useful.

`ThriftType.UserType` should subsume `Named`, and `ThriftType` in general should become enriched.  This would fix all the fiddly `Constant` validation bugs and simplify lots of requested use cases from @hzsweers and friends.

benjamin-bader:
So - the proposed hierarchy would be:
```
ThriftType
|__BuiltinThriftType (bool,byte,i8,i16,i32,i64,double,string,binary)
|__ListThriftType
|__SetThriftType
|__MapThriftType
|__TypedefThriftType
|__UserThriftType
|  |__EnumThriftType
|  |__StructThriftType (struct, union, exception)
|__ServiceThriftType
```

`UserThriftType` and `ServiceThriftType`, and `Constant` will tentatively implement an interface exposing the same things that `Named` does currently - `Location` and Javadoc.

`Program` would be more or less the same, but everything from `Linker` down changes.  Codegen will presumably be different, though I haven't thought much about it.

hzsweers:
After what we discussed about Typedefs, should they maybe straddle the line with UserThriftType a bit more?

benjamin-bader:
You and I arrived at the same conclusion; I've ended up with both TypedefType and ServiceType extending UserType.  The final hierarchy, then, is

```
ThriftType
|__BuiltinType (bool,byte,i8,i16,i32,i64,double,string,binary)
|__ListType
|__SetType
|__MapType
|__UserType
|  |__EnumType
|  |__StructType (struct, union, exception)
|  |__TypedefType
|  |__ServiceType
```

PR #71, opened 2016-10-24
Fix builders by including Nullable fields contained in equals().

naturalwarren:
This pull request adds fields to builders that are present in equals(). Updated unit tests to assert fields added to Builders. Thanks to Zac for pointing this out. 

Zac also mentioned all namespace builders are taking Map instead of ImmutableMap. They should either take ImmutableMap or make a defensive copy. Ben, is there a preference here?

### DIFF
--- a/thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Field.java
+++ b/thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Field.java
@@ -54,6 +54,7 @@ public final class Field {
         this.fieldNamingPolicy = builder.fieldNamingPolicy;
         this.annotations = builder.annotations;
         this.type = builder.type;
+        this.javaName = builder.javaName;
     }
 
     public Location location() {
@@ -128,7 +129,7 @@ public final class Field {
     }
 
     public Builder toBuilder() {
-        return new Builder(element, fieldNamingPolicy, annotations, type);
+        return new Builder(element, fieldNamingPolicy, annotations, type, javaName);
     }
 
     public static final class Builder {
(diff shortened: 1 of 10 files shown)


thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Field.java
```diff
@@ -54,6 +54,7 @@ private Field(Builder builder) {
         this.fieldNamingPolicy = builder.fieldNamingPolicy;
         this.annotations = builder.annotations;
         this.type = builder.type;
+        this.javaName = builder.javaName;
```
benjamin-bader: I wouldn't recommend adding `javaName` to builders; it's a computed field, determined by the thrift name and the naming policy.

In fact, this field is going to go away soon as part of the work on #69; probably not worth spending much time on it.

benjamin-bader:
Regarding maps vs immutable maps, I don't have a strong preference; I guess it depends on how the builders will work.  I'd say stick with a defensive copy of a map if:
- we want to expose a builder API like `Builder withNamespace(NamespaceScope, String)`
- or, if the other collections in the builder are also mutable

If not, we ought to just be consistent with the rest of the builder params and use immutable maps.

thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Field.java
hzsweers: Should we maybe mark it as transient while we're here for posterity or just leave it as is since it's going away?

hzsweers:
Maps with a defensive copy might be the nicest in the API though, then maybe a check under the hood to just keep the parameter if it's an immutablemap instance

benjamin-bader:
Mutable maps definitely give more options for nice APIs; if we want to make nice builder APIs, let's keep them.

naturalwarren:
Awesome thanks Ben and Zac.

naturalwarren:
Is there anything else you'd like me to do here? I'm not sure how I can resolve the conflicts. I'm currently rebased on top of master.

benjamin-bader:
Yeah, I'm not sure why there is a conflict, but it's minor.  I suspect that you could manually squash your commits then `git push -f origin ws_builder_fix`, and the problem would be resolved.

Alternately, I could do the merge manually in the same way; the drawback is that your name would no longer be on the commit.

benjamin-bader:
Also, it's worth checking, is your `master` up-to-date with `upstream/master`?

hzsweers:
@naturalwarren if your setup is still the same from when we set it up awhile back, should be:

```
git checkout master
git fetch upstream
git merge upstream/master
git checkout ws_builder_fix
git rebase master
git push -f origin ws_builder_fix
```

naturalwarren:
Thanks guys. That did it!

benjamin-bader:
Thanks, @naturalwarren!

PR #72, opened 2016-10-25
Keep namespaces in ThriftType.TypedefType for later reference

hzsweers:
This is useful in cases where you use Typedefs in generated code directly for type safety. Verified with a local install and worked swimmingly 👍 . Made easier by the fact that the base type was already holding onto the namespace, we just need to pass it on to the typedef creation.


hzsweers:
@benjamin-bader let me know if/how you want me to test this, as I didn't see anything directly around namespaces

hzsweers:
Something else to consider: The TypeResolver implementation could adopt the same logic as visitUserType and allow for java namespace retrieval if you wanted

benjamin-bader:
I don't think we'll want to have TypeResolver use this new map, since it is specific to our current code gen.  That aside, this PR looks great.  Thanks!

hzsweers:
Thanks!

PR #73, opened 2016-10-29
Refactor thrift type

benjamin-bader:
This major refactoring fixes #66 and #69.

As discussed in the latter, `Named` goes away, and is merged in with `ThriftType`.  `ThriftType` is a branchy hierarchy, and many of its elements share common structure that isn't easily expressed via direct inheritance.  To that end, the new `UserElement` interface captures that commonality, and `UserElementMixin` encapsulates the logic; various types and components can implement the former by containing the latter.

Builders also get a proper refactoring with the introduction of inheritance and lots of generics garbage to make a nice API.

This is a pretty big change, though (hopefully) transparent to consumers; I'd appreciate eyeballs or, even better, testing from @hzsweers, @naturalwarren , @seanabraham, and any other folks who see this and are interested.

PR #74, opened 2016-10-30
Add withNamespaces() API to ThriftType

hzsweers:
This allows you to construct a new instance with modified namespaces, and also fixes a bug in the merge() logic where it assumed putAll() overwrote existing keys (it actually errors).

Would be great to get this in or pulled in to the work in #73, just putting here for reference in case.


benjamin-bader:
With the proviso that this code will be going away with #73, I'm fine with this PR.  Thanks!

hzsweers:
Thanks!

PR #98, opened 2017-02-10
Add convenience type() to user element builder

hzsweers:
This lets you do the update from an existing type in a batch update

### DIFF
--- a/thrifty-schema/src/main/java/com/microsoft/thrifty/schema/AbstractUserElementBuilder.java
+++ b/thrifty-schema/src/main/java/com/microsoft/thrifty/schema/AbstractUserElementBuilder.java
@@ -34,6 +34,7 @@ import java.util.Map;
 abstract class AbstractUserElementBuilder<
         TElement extends UserElement,
         TBuilder extends AbstractUserElementBuilder<TElement, TBuilder>> {
+
     UserElementMixin mixin;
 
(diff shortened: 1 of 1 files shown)


naturalwarren:
Looks good.

thrifty-schema/src/main/java/com/microsoft/thrifty/schema/AbstractUserElementBuilder.java
```diff
@@ -64,5 +65,16 @@ public TBuilder annotations(Map<String, String> annotations) {
         return (TBuilder) this;
     }
 
+    public TBuilder type(TElement type) {
```
benjamin-bader: Is this intended to be a "set all builder params to this `TElement`" kind of function, or do you imagine it to be just constrained to `UserElement` values?  As is, I like it but can imagine surprises down the road:

```
StructType s1 = ...
StructType s2 = ... //something totally different

s1.toBuilder().
    type(s2)  // why?  who knows!  but now the builder is basically s2 with the fields of s1
```

hzsweers: Agreed it could be misused, but then again so could the others I think (especially since this is just a convenience setter over existing ones)

PR #99, opened 2017-02-10
Add builders for servicemethod and constant

hzsweers:
This also adds some convenience toBuilders() on their respective elements

### DIFF
--- a/thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Constant.java
+++ b/thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Constant.java
@@ -22,15 +22,14 @@ package com.microsoft.thrifty.schema;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
-import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.microsoft.thrifty.schema.parser.ConstElement;
 import com.microsoft.thrifty.schema.parser.ConstValueElement;
-import com.microsoft.thrifty.schema.parser.FunctionElement;
 
-import javax.annotation.Nullable;
 import java.util.List;
 import java.util.Map;
+
+import javax.annotation.Nullable;
 
 /**
  * Represents a Thrift const definition.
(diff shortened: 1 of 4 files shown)


naturalwarren:
Would be a good idea to add equals() and hashcode() with some unit tests.

thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Constant.java
```diff
@@ -104,12 +112,73 @@ void validate(Linker linker) {
...
+        return new Builder(this);
+    }
+
     @VisibleForTesting
     static void validate(Linker linker, ConstValueElement value, ThriftType expected) {
         ThriftType trueType = expected.getTrueType();
         Validators.forType(trueType).validate(linker, trueType, value);
     }
 
+    public static final class Builder {
```
benjamin-bader: I think this could be an `AbstractUserElementBuilder`, no?

hzsweers: I thought that required one parameter to extend usertype. Can check again

thrifty-schema/src/main/java/com/microsoft/thrifty/schema/ServiceMethod.java
```diff
@@ -153,4 +168,75 @@ void validate(Linker linker) {
             linker.addError(field.location(), "Only exception types can be thrown");
         }
     }
+
+    public static final class Builder {
```
benjamin-bader: This could also be an `AbstractUserElementBuilder`

thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Constant.java
benjamin-bader: `UserElement`, the interface implemented by the mixin.

thrifty-schema/src/main/java/com/microsoft/thrifty/schema/ServiceMethod.java
hzsweers: 7abce2c

thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Constant.java
hzsweers: 7abce2c

PR #100, opened 2017-02-10
More lenient collection types in builders

hzsweers:
AutoValue can detect guava's immutables and do the defensive copy under the hood for us


naturalwarren:
Looks good.

PR #101, opened 2017-02-10
Replace annotations field in UserElementMixin directly rather than building an additive map

hzsweers:
Matches conventions in the other builders


benjamin-bader:
Thanks!

PR #103, opened 2017-02-15
Fix enum-member parsing with doc comments

benjamin-bader:
Fixes an issue reported out-of-band by @hzsweers

### DIFF
--- a/thrifty-schema/src/test/java/com/microsoft/thrifty/schema/parser/ThriftParserTest.java
+++ b/thrifty-schema/src/test/java/com/microsoft/thrifty/schema/parser/ThriftParserTest.java
@@ -1124,6 +1124,45 @@ public class ThriftParserTest {
     }
 
     @Test
+    public void enumJavadocWithoutSeparators() {
+        String thrift = "/**\n" +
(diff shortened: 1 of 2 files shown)


thrifty-schema/src/test/java/com/microsoft/thrifty/schema/parser/ThriftParserTest.java
```diff
@@ -1124,6 +1124,45 @@ public void trailingDocWithoutSeparatorWithAnnotationOnNewLine() {
     }
 
     @Test
+    public void enumJavadocWithoutSeparators() {
+        String thrift = "/**\n" +
+                " * Some Javadoc\n" +
+                " */\n" +
+                "enum Value {\n" +
```
hzsweers: Did this test fail before this change? I couldn't actually get a repro case :O

benjamin-bader: Yes, it did - this is basically the snippet you shared, minus one member and with things renamed.  Not sure why it wouldn't have failed for you.

Issue #145, opened 2017-05-29
Synchronous requests


gotev:
It will be useful to be able to make also synchronous Thrift requests, because they are useful inside Android intent services. Currently to make a call to `myFunction`, the only way is:
```java
client.myFunction(request, new ServiceMethodCallback<MyResponse>() {
    @Override
    public void onSuccess(MyResponse result) {
        Log.i("Response", result.accessToken);
    }

    @Override
    public void onError(Throwable error) {
        Log.e("Error", "Error", error);
    }
});
```
Which is async and not suitable to be used inside a service as is.

By [wrapping the callback inside a Future](https://stackoverflow.com/a/2180534), I made a workaround to make synchronous calls, although the resulting code is not as easy to read and generates some overhead as well.
```java
try {
    SyncServiceMethodCallback<MyResponse> responseFuture = new SyncServiceMethodCallback<>();

    client.myFunction(request, responseFuture);

    MyResponse result = responseFuture.get();

    Log.i("Response", result.accessToken);

} catch (Throwable error) {
    Log.e("Error", "Error", error);
}
```
The ideal will be to have:
```java
try {
    MyResponse result = client.myFunction(request);

    Log.i("Response", result.accessToken);

} catch (Throwable error) {
    Log.e("Error", "Error", error);
}
```
I would like to contribute to this great project, but I wanted to know which is the best way to make such implementation in your opinion, based on the current library design. Something like Retrofit's CallAdapterFactory will be great, so other adapters could be easily developed as plugins (e.g. Rx).

Btw, here there's `SyncServiceMethodCallback` code, just in case someone else needs it:
```java
import android.support.annotation.NonNull;
import com.microsoft.thrifty.service.ServiceMethodCallback;
import java.util.concurrent.CountDownLatch;
import java.util.concurrent.ExecutionException;
import java.util.concurrent.Future;
import java.util.concurrent.TimeUnit;
import java.util.concurrent.TimeoutException;

public class SyncServiceMethodCallback<T> implements ServiceMethodCallback<T>, Future<T> {

    private volatile T callbackResult = null;
    private volatile boolean cancelled = false;
    private volatile Throwable exception = null;
    private final CountDownLatch countDownLatch;

    public SyncServiceMethodCallback() {
        countDownLatch = new CountDownLatch(1);
    }

    @Override
    public void onSuccess(T result) {
        callbackResult = result;
        countDownLatch.countDown();
    }

    @Override
    public void onError(Throwable error) {
[...]

PR #147, opened 2017-06-18
Split ClientBase to enabled synchronous calls

gabrielittner:
This is my first go at splitting `ClientBase` for #145 

The second commit contains the actual changes, the other two are just make the diff of the second nicer.

### DIFF
--- a/thrifty-runtime/src/main/java/com/microsoft/thrifty/service/AsyncClientBase.java
+++ b/thrifty-runtime/src/main/java/com/microsoft/thrifty/service/AsyncClientBase.java
@@ -88,13 +88,11 @@ public class AsyncClientBase extends ClientBase implements Closeable {
      */
     private final BlockingQueue<MethodCall<?>> pendingCalls = new LinkedBlockingQueue<>();
 
-    private final Protocol protocol;
     private final Listener listener;
     private final WorkerThread workerThread;
 
     protected AsyncClientBase(Protocol protocol, Listener listener) {
         super(protocol);
-        this.protocol = protocol;
         this.listener = listener;
         this.workerThread = new WorkerThread();
 
@@ -131,11 +129,7 @@ public class AsyncClientBase extends ClientBase implements Closeable {
 
         workerThread.interrupt();
 
-        try {
-            protocol.close();
-        } catch (IOException ignored) {
-            // nope
-        }
+        closeProtocol();
 
         if (!pendingCalls.isEmpty()) {
             List<MethodCall<?>> incompleteCalls = new ArrayList<>();
--- a/thrifty-runtime/src/main/java/com/microsoft/thrifty/service/AsyncClientBase.java
+++ b/thrifty-runtime/src/main/java/com/microsoft/thrifty/service/AsyncClientBase.java
@@ -208,16 +208,15 @@ public class AsyncClientBase extends ClientBase implements Closeable {
             Exception error = null;
             try {
                 result = AsyncClientBase.this.invokeRequest(call);
-            //TODO a nicer way would be a struct exception base class and only catch that
             } catch (IOException | RuntimeException e) {
                 throw e;
             } catch (Exception e) {
                 if (e instanceof Struct) {
                     error = e;
-                } else {
-                    // invokeRequest should only throw IOException, RuntimeExceptions
+                } else if (!(e instanceof AbortException)) {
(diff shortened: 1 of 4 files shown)


thrifty-runtime/src/main/java/com/microsoft/thrifty/service/AsyncClientBase.java
```diff
@@ -0,0 +1,267 @@
...
+
+            if (call == null) {
+                return;
+            }
+
+            Object result = null;
+            Exception error = null;
+            try {
+                result = AsyncClientBase.this.invokeRequest(call);
+            //TODO a nicer way would be a struct exception base class and only catch that
```
gabrielittner: See todo. Do we want something like a `StructException` as base class for the explicitly declared exceptions which are passed to `fail()`?

thrifty-runtime/src/main/java/com/microsoft/thrifty/service/ClientBase.java
```diff
@@ -85,226 +48,96 @@
...
-                throw new ThriftException(
-                        ThriftException.Kind.WRONG_METHOD_NAME,
-                        "Unexpected method name in reply; expected " + call.name
-                                + " but received " + metadata.name);
-            }
+        if (metadata.type == TMessageType.EXCEPTION) {
+            ThriftException e = ThriftException.read(protocol);
+            protocol.readMessageEnd();
+            handleExceptionMessage(call, e);
+            //TODO there was no throw/return here before but the sync impl of handleExceptionMessage throws
```
gabrielittner: Not sure about this one. It used to call fail, but not return. Does this mean we can get a proper result or a second fail call afterwards?

thrifty-runtime/src/main/java/com/microsoft/thrifty/service/ClientBase.java
```diff
@@ -85,226 +48,96 @@
...
     /**
-     * When invoked by a derived instance, places the given call in a queue to
-     * be sent to the server.
+     * When invoked by a derived instance, sends the given call to the server.
      *
      * @param methodCall the remote method call to be invoked
+     * @return the result of the method call
      */
-    protected void enqueue(MethodCall<?> methodCall) {
+    protected final Object execute(MethodCall<?> methodCall) throws Exception {
```
gabrielittner: Do we want this here or a second `ClientBase` subclass like `SyncClientBase` (in that case `handleExceptionMessage()` could be abstract in here).

benjamin-bader:
Thanks for the contribution - this looks pretty good!  I'd like to take a little bit to review, but have some other work I have to handle first.  I'm excited to get this in, so thanks for your patience in the meantime!

thrifty-runtime/src/main/java/com/microsoft/thrifty/service/ClientBase.java
benjamin-bader: Hm, if it didn't bail out in the old version, that's definitely a bug.  A `ThriftException` usually means that the client and server will never be able to talk successfully to one another!

Throwing in the sync client is the right thing to do; async should finish reading the message and return.  Good catch!

thrifty-runtime/src/main/java/com/microsoft/thrifty/service/ClientBase.java
benjamin-bader: Mmm, I could go either way, but this seems simpler for now.  It would be an easy refactor later, if we find that we need the separation.

thrifty-runtime/src/main/java/com/microsoft/thrifty/service/ClientBase.java
gabrielittner: I've changed this to throw a custom exception after `handleExceptionMessage()`. For sync that will never be reached because the `ThriftException` is thrown before. The async client will catch the custom exception and just swallow it because it's `handleExceptionMessage()` already called `fail()`.  (Kotlin's `Nothing` as return type for `handleExceptionMessage()` would be really nice here)

We should probably add a test that the async client doesn't call `fail()` twice after receiving a `TMessageType.EXCEPTION`. I'm not very familiar with server thrift code, is there an easy way to make it send one?

benjamin-bader:
Awesome, thank you!

PR #150, opened 2017-07-16
Fix server error handling

gabrielittner:
The current implementation of `AsyncClientBase` will call success after for `TMessageType.EXCEPTION` exceptions. The simple fix would have been to return  when getting an `AbortException`, but I decided to streamline it a bit. Now fail and success are each only called in a single place and there is no separate exception handling anymore.

### DIFF
--- a/thrifty-runtime/src/main/java/com/microsoft/thrifty/service/ClientBase.java
+++ b/thrifty-runtime/src/main/java/com/microsoft/thrifty/service/ClientBase.java
@@ -69,7 +69,7 @@ public class ClientBase implements Closeable {
 
         try {
             return invokeRequest(methodCall);
-        } catch (UndeclaredServerException e) {
+        } catch (ServerException e) {
             throw e.thriftException;
         }
     }
@@ -100,8 +100,8 @@ public class ClientBase implements Closeable {
      *
      * @param call the remote method call to be invoked
      * @return the result of the method call
+     * @throws ServerException wrapper around {@link ThriftException}. Callers should catch and unwrap this.
      * @throws IOException from the protocol
-     * @throws UndeclaredServerException exception received from server that was not declared in schema
      * @throws Exception exception received from server implements {@link com.microsoft.thrifty.Struct}
      */
     final Object invokeRequest(MethodCall<?> call) throws Exception {
@@ -129,7 +129,7 @@ public class ClientBase implements Closeable {
         if (metadata.type == TMessageType.EXCEPTION) {
             ThriftException e = ThriftException.read(protocol);
             protocol.readMessageEnd();
-            throw new UndeclaredServerException(e);
+            throw new ServerException(e);
         } else if (metadata.type != TMessageType.REPLY) {
             throw new ThriftException(
                     ThriftException.Kind.INVALID_MESSAGE_TYPE,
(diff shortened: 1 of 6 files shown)


thrifty-runtime/src/main/java/com/microsoft/thrifty/service/ClientBase.java
```diff
@@ -67,7 +67,11 @@ protected final Object execute(MethodCall<?> methodCall) throws Exception {
             throw new IllegalStateException("Cannot write to a closed service client");
         }
 
-        return invokeRequest(methodCall);
+        try {
+            return invokeRequest(methodCall);
+        } catch (UndeclaredServerException e) {
+            throw e.thriftException;
```
gabrielittner: This maintains the old behavior of `handleExceptionMessage` which just threw the exception. I'm not sure if that is really what we want for synchronous clients. We have the exception wrapping explicitly so that we can pass it on to `fail()` for asynchronous call. That way the client doesn't get closed and we can continue to execute different calls. 
Should be expose `UndeclaredServerException` as public API so that a synchronous call has to catch it? Throwing a runtime exception is pretty similar to closing the async client and catching `ThriftException` is probaby not a good idea because in other cases you want it to go through.

benjamin-bader:
`ThriftException` usually means that client and server are incompatible - they aren't using the same service definitions, or the client is trying to connect to a non-thrift endpoint, etc.  To me, it has always seemed better to just throw it and let someone else decide how to handle it.

Why would we not want that for synchronous clients?

gabrielittner:
I'm not suggesting to handle the exception inside thrifty for the synchronous client. The current behavior of the sync client (both before and after this pr) is that all ThriftExceptions are thrown. The question is do we want to make it easier for consumers to handle this case of ThriftException on a per method level. The async client does that by passing it to `onError()` in `ServiceMethodCallback`. Other ThriftExceptions are thrown, cause a client shutdown and are than passed to `AsyncClientBase.Listener.onErrror()`. The sync client does not have different callbacks so we could use a different type to do that.

gabrielittner:
>  this case of ThriftException
I'm referring to ThriftExceptions caused by a response with `metadata.type == TMessageType.EXCEPTION`. They seem to happen when the server throws a `TException` that isn't declared in the schema (not sure if there are other cases where).

benjamin-bader:
As far as I recall, `TMessageType.EXCEPTION` is sent in response to protocol-level problems, as well as undeclared server exceptions.  I actually forgot about the undeclared-server-exception behavior!  In that light, I agree that we should just pass all of these on to client callbacks, rather than throw some and pass others.

It seems like the Apache implementation does it that way, too - any `TMessageType.EXCEPTION` in an async client gets passed to the method's callback, as opposed to shutting down the client.

thrifty-runtime/src/main/java/com/microsoft/thrifty/service/ClientBase.java
benjamin-bader: I probably wouldn't want to expose it with the name `UndeclaredServerException`, but I can see the value in having a checked exception on synchronous client methods.  What do you think about making `ThriftException` checked?

gabrielittner: Sounds good to me, I'll update the PR tomorrow.

gabrielittner:
Should be good to go now.

benjamin-bader:
Hm, I didn't realize that this change would require making adapters declare a checked exception! I see that the need arises from the enum validation that we do in `GenerateServiceReader`.  I think that we should land this change, then I'll switch the generated exception to a RuntimeException of some sort and undo just the adapter change.

Thanks for bearing with me!  Now, on to your larger change...

PR #151, opened 2017-07-17
Generate synchronous clients

gabrielittner:
This is my first draft to generate synchronous clients for #145 

Open questions and todos:
- [ ] Do we want to support having synchronous and asynchronous clients side by side (current implementation)? It definitely helps with the integration tests.
- [ ] Naming of synchronous clients: Currently `Sync` is added to the interfaces and `SyncClient` to the classes. If we only support having one client type that can be removed. If not are there better suffixes?
- [ ] `ServiceBuilder` is still pretty close to the original one, not sure if the current base class would be enough to implement a rx client on top of it
- [ ] improve compiler flags, document them
- [ ] clean up some parts of the code

- [x] generated catch block doesn't use imports

Here is what the generated synchronous client looks like: https://gist.github.com/gabrielittner/55e91c27b34d64e152e8cabee5cc2b5d


gabrielittner:
Rebased and one new commit

gabrielittner:
@benjamin-bader Did you get a chance to look at this?

benjamin-bader:
Very sorry to have let this sit for so long - I haven't been a good maintainer here.  I should have responded here a year ago.  I was embarrassed both to reject a PR after having given such specific guidance, and to have wasted your time, but I was wrong not to have spoken up.

On reading the code, I wasn't happy with the increasingly-complex direction the codegen was taking.  It felt (and still feels) as if making `AsyncServiceBuilder` extend `ServiceBuilder` was the wrong approach - the code reuse is a false economy, because the methods being overridden is so specific to the implementation.  It's not so advisable to have "interface" methods like `getServiceInterfaceSuffix()`.  Not only are they too specific to really constitute a useful interface, they render the base code quite brittle.

After seeing the code, it became apparent that a better approach would be to decouple the two implementations - ideally, also collapsing the `ClientBase`/`AsyncClientBase` split.  For that reason, it would be better not to pursue this approach.

PR #152, opened 2017-08-06
Add UUID to parsing for matching elements across instances

hzsweers:
This is a prototype of trying to tag parsed elements with unique IDs so that one could match them even if you `newBuilder`'d it, you could match back other elements that reference the now-old reference to them. This would allow for one to normalize references in a `Schema` after changing a bunch of elements in it, such as with a preprocessor to edit Schemas before code gen.

If the implementation here looks good, I can proceed with figuring out how to update tests to handle this in a deterministic way

### DIFF
--- a/thrifty-schema/src/main/java/com/microsoft/thrifty/schema/parser/ThriftyParserPlugins.java
+++ b/thrifty-schema/src/main/java/com/microsoft/thrifty/schema/parser/ThriftyParserPlugins.java
@@ -1,0 +1,72 @@
+package com.microsoft.thrifty.schema.parser;
+
+import java.util.UUID;
+
+/**
+ * Utility class to inject handlers to certain standard Thrifty operations.
+ */
+public final class ThriftyParserPlugins {
+
+    private static final UUIDProvider DEFAULT_UUID_PROVIDER = UUID::randomUUID;
+    private static volatile UUIDProvider uuidProvider = DEFAULT_UUID_PROVIDER;
+
+    /**
+     * Prevents changing the plugins.
+     */
+    private static volatile boolean lockdown;
+
+    /**
+     * Prevents changing the plugins from then on.
+     * <p>
+     * This allows container-like environments to prevent client messing with plugins.
+     */
+    public static void lockdown() {
+        lockdown = true;
+    }
+
+    /**
+     * Returns true if the plugins were locked down.
+     *
+     * @return true if the plugins were locked down
+     */
+    public static boolean isLockdown() {
+        return lockdown;
+    }
+
+    /**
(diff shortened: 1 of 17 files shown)


hzsweers:
@benjamin-bader absolutely no CR expected for at least 3-4 weeks! Focus on your honeymoon :)

benjamin-bader:
Hm, the tests failed, but it's not immediately clear why.  I'm guessing that the tests haven't been updated to account for the new UUID member, probably the AutoValue parser elements that would automatically include it in `.equals()`.

hzsweers:
Yeah I haven't updated the tests yet, wanted to get a review of the API first before taking a crack at them

benjamin-bader:
API looks fine to me; it's what I'd do if I were implementing the feature, so no complaints here.

hzsweers:
Cool, I'll work on updating the tests then. My initial thinking is to basically add some way to set a seed to use for uuids for test-only purposes. Alternatively, we could make mark the uuid as transient to autovalue somehow so it's ignored, but not sure how nicely that plays with builders and whatnot. Will try some stuff

thrifty-schema/src/main/java/com/microsoft/thrifty/schema/parser/ThriftyParserPlugins.java
```diff
@@ -0,0 +1,72 @@
+package com.microsoft.thrifty.schema.parser;
+
+import java.util.UUID;
+
+/**
+ * Utility class to inject handlers to certain standard Thrifty operations.
+ */
+public final class ThriftyParserPlugins {
```
benjamin-bader: Interesting, I must have missed this class the last time I looked at this PR.  Why would you need something other than `UUID::randomUUID` - for testing, it seems?

If so, and there aren't any larger plans for this class, have you considered simply adding a `UUIDProvider` param to a `ThriftParser` constructor, and making the parser itself responsible for assigning the UUIDs?  It would mean also changing the autovalue factory functions for the parser elements, to add a UUID, but would have the benefit of being less stateful.

I think you might have deliberately decided against doing this, so I'm curious about why.

hzsweers: I added it in this followup. The primary case is testing, but one could conceivably also use this to supply their own no-op version if they don't need it and want the slight perf boost from not calculating a random one.

I opted to make it a plugin setup because I wanted to have a simple static location for both simplicity in the parser's access as well as making it public for anyone creating/modifying their own types to use if they create their own (to keep things consistent)

benjamin-bader:
Thanks!  For the PR, and for your patience seeing it through.

hzsweers:
Thanks!

PR #153, opened 2017-08-09
Fix reading bool value using compact protocol

reisub:
We came across an issue in deserializing a bool value using compact protocol - the value always comes out false.

This PR modifies the test so that it catches this bug and fixes it.

### DIFF
--- a/thrifty-integration-tests/ClientThriftTest.thrift
+++ b/thrifty-integration-tests/ClientThriftTest.thrift
@@ -66,7 +66,9 @@ struct Xtruct
   1:  string string_thing,
   4:  byte   byte_thing,
   9:  i32    i32_thing,
-  11: i64    i64_thing
+  11: i64    i64_thing,
+  13: double double_thing,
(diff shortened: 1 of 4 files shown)


benjamin-bader:
Excellent catch - thanks for the fix!  Would you mind including the thrift change you made to Xtruct as well, so we don't lose it if/when we ever regenerate it?

reisub:
@benjamin-bader done in https://github.com/Microsoft/thrifty/pull/153/commits/8840ab3eef372bddb5fb22fcac0170f2df11de0c, though I'm not sure which exact file you wanted changed since there are none in the `thrift-runtime` module.

benjamin-bader:
Perfect.  There's no one file, sadly - we seem to have overlooked automating generating this file, probably because it is done so infrequently.  I think this would only be the second version.

I'm comfortable calling the integration-test file the "source of truth" here.

benjamin-bader:
Merged - thanks again!

PR #156, opened 2017-10-16
Json Protocol support

### LINKED ISSUE #154: Json Protocol support
Would you be open to a PR implementing json protocol support, based on Apache Thrift implementation? This functionality would be quite useful for easy request/response logging in development.

reisub:
I started from the official Thrift implementation, made it work and tried to clean up the style in the process. Let me know if there's anything you'd like me to change, happy to do it.

I already signed the contributor agreement for an earlier PR.

Closes #154.

### DIFF
--- a/thrifty-runtime/src/test/java/com/microsoft/thrifty/protocol/JsonProtocolTest.java
+++ b/thrifty-runtime/src/test/java/com/microsoft/thrifty/protocol/JsonProtocolTest.java
@@ -1,0 +1,195 @@
+/*
+ * Thrifty
+ *
+ * Copyright (c) Microsoft Corporation
+ *
+ * All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the License);
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * THIS CODE IS PROVIDED ON AN  *AS IS* BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING
+ * WITHOUT LIMITATION ANY IMPLIED WARRANTIES OR CONDITIONS OF TITLE,
+ * FITNESS FOR A PARTICULAR PURPOSE, MERCHANTABLITY OR NON-INFRINGEMENT.
+ *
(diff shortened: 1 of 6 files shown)


reisub:
@benjamin-bader did you have time to look at this code?

thrifty-runtime/src/test/java/com/microsoft/thrifty/protocol/JsonProtocolTest.java
```diff
@@ -0,0 +1,195 @@
...
+                "\"4\":{\"i8\":1}," +
+                "\"9\":{\"i32\":3}," +
+                "\"11\":{\"i64\":4}," +
+                "\"13\":{\"dbl\":2.0}" +
+                "}");
+    }
+
+    @Test
+    public void binary() throws Exception {
+        protocol.writeBinary(ByteString.encodeUtf8("foobar"));
```
benjamin-bader: I think we should exercise the base-64 coder and decoder more thoroughly here.  Let's generate an array of bytes from 0..255 and encode that, so that we hit all of the encoding/decoding table entries.

I'm assuming that those are copied verbatim from Apache, and so are almost certainly correct, but we shouldn't assume such things in tests :)

benjamin-bader:
Yes, sorry for the delay - I've been an overloaded maintainer on quite a few projects for the past while.  This change is nice and self-contained, which is great, but is also large.  Am I correct in thinking that this is mainly a copy-and-adapt implementation based on Apache's?

thrifty-runtime/src/test/java/com/microsoft/thrifty/protocol/JsonProtocolTest.java
reisub: Yes, your assumption is correct. And yeah, I'll add tests for the encoding and decoding.

reisub:
@benjamin-bader I understand - don't worry about it, it's not urgent.

You're right, Apache's implementation was the basis for this PR.

reisub:
I wrote a test for the base64 decoding (see https://github.com/Microsoft/thrifty/pull/156/commits/e33a66da0eeb7a77169f060481fbe3f2db6bf342).

But then I realized we can use the base64 decoding from the Okio `ByteString` class. So I changed the implementation to use that instead, making it simpler.

@benjamin-bader let me know if you want anything else changed.

reisub:
@benjamin-bader is there anything I can do to help you review and merge this work?

benjamin-bader:
Thanks for contributing!

PR #157, opened 2017-11-22
Update Gradle to 4.3.1

jaredsburrows:
@benjamin-bader


benjamin-bader:
Looks like this fails because it brings in a new version of Checkstyle, from which the `RedundantThrows` module was removed.  We should remove that module from `config/checkstyle/checkstyle.xml`, which fixes the break.

jaredsburrows:
@benjamin-bader Done!

benjamin-bader:
Thanks, Jared!

Issue #160, opened 2018-01-27
Generate `Nullable` for all fields in a union


bangarharshit:
Currently, thrifty checks that ` a number of fields set= 1` for unions but doesn't annotate the fields with Nullable. 

It can be a useful indication for devs as well as tools like infer/NullAway. We can use this blog post if the resulting code is too cluttered  - https://medium.com/square-corner-blog/non-null-is-the-default-58ffc0bb9111

benjamin-bader:
Adding nullable to union fields would definitely be an improvement!  I'd be happy to accept a PR - it would be quite simple to modify ThriftyCodeGenerator#buildStruct to do this.

bangarharshit:
@benjamin-bader added PR for the union. 

Do you also want to support non-null by default? It may reduce the clutter but not sure how useful will it be for generated code. All it needs is to get rid of NonNull annotation and create a package-info.java for all the generated packages.

benjamin-bader:
Thanks for the PR!  I just realized that we already did emit nullable annotations for unions, but with your change, that's now explicit - our codebase is more robust thanks to it.

I don't think nonnull-by-default is worth worrying about - these annotations don't do much to obscure the generated code (which, after all, is _generated_ code).  We value its readability, but only to a certain point.  Is there a use-case here, beyond producing more aesthetically-pleasing code?

bangarharshit:
None, Apart from lesser boilerplate. 

TBH, I agree that for the generated code it is ok to use @NonNull everywhere and adds to the complexity. Most of the static code analysis tools such as infer and nullaway supports both. Closing the ticket. Feel free to re-open the ticket if you feel it is worth it.

Thanks.

PR #161, opened 2018-01-28
Generated nullable annotations for all the fields in union

bangarharshit:
Fixes - https://github.com/Microsoft/thrifty/issues/160. @benjamin-bader do you think we should throw an error if a union is annotated with anything such as required or optional? [Thrift doc](https://thrift.apache.org/docs/idl#union) suggests that union's default value is optional. Does this mean union can be required? It kind of breaks the semantics.


benjamin-bader:
Thanks for the PR - looks good!

benjamin-bader:
Re: throwing if unions have `required`, we do that as part of the schema validation in the `thrifty-schema` module: 
https://github.com/Microsoft/thrifty/blob/master/thrifty-schema/src/main/java/com/microsoft/thrifty/schema/StructType.java#L109

Issue #162, opened 2018-02-03
Nullability friendly builder


bangarharshit:
Most of the static code analysis tools such as NullAway and infer throws error with the current builder setup.

There are 2 issues:
1. Builder fields are not annotated.
2. Builder fields can't be non-null with the current setup since the constructor is usually empty.

```java
public Builder() {}

public Builder nonNullField(@NonNull String nonNullField) {}

public Builder nullableField(@Nullable String nullableField) {}
```
 An alternate way will be to set all the required values in Builder's constructor and only have separate methods for Nullable fields.

```java
public Builder(@NonNull String nonNullField) {}

public Builder nullableField(@Nullable String nullableField) {}
```

Since this breaks backward compatibility it needed to be flagged.

Complete class - https://gist.github.com/bangarharshit/c4ec0bee365cc4ef30270f10d4d1aa2f

benjamin-bader:
I think annotating parameters of setter methods is a great idea - I'm surprised we don't already do that!  I'm not familiar with errors from infer or nullaway - which parts of the generated code get flagged?

I'd rather not force builders to be "correct" on construction; they are intended to be a readable way to build up their companion struct, and stuffing all the required params in their constructor would mostly defeat that purpose.

bangarharshit:
That is a very good point. Here is a demo [project](https://github.com/bangarharshit/Nullness-Demo/tree/thrity) with readme containing all the error details. 

An alternative will be to use some kind of `required` annotation on top of builder methods. Let me surface that to these checker teams.

Adding annotation to both parameters and fields will still be a very useful.

benjamin-bader:
Thanks for the demo project!  I see that NullAway indeed doesn't like the builder fields being without annotations.

I think it's interesting that there's no warning in the `Arbit` constructor - looks like the tool is smart enough to see that the ctor is manually checking for null and throwing?  In that case, how about just annotating every builder field as `@Nullable`, and annotating the setter parameters according to the field's requiredness?  e.g.

```java
class ArbitBuilder {
  @Nullable String requiredField = null;
  @Nullable String optionalField = null;

  ArbitBuilder setRequiredField(@NonNull String required) {
    ...
  }

  ArbitBuilder setOptionalField(@Nullable String optionalField) {
    ...
  }
}
```

This way, the tool would be satisfied, and the annotations themselves would still be honest.  Does that sound like it would work?  What do you think?

bangarharshit:
[...]

PR #164, opened 2018-02-05
Fixed nullability annotation for static code analysis tools

bangarharshit:
Fixes - https://github.com/Microsoft/thrifty/issues/162. 
There are 2 changes:
1. The nullability check for the field is moved from Builder's build method to constructor since we are enforcing nullness on object's field.
Current code:
```java
    @Override
    public Email build() {
      if (this.From == null) {
        throw new IllegalStateException("Required field 'From' is missing");
      }
      if (this.Attachments == null) {
        throw new IllegalStateException("Required field 'Attachments' is missing");
      }
      return new Email(this);
    }
```
Proposed code:
```java
  private Arbit(Builder builder) {
    if (builder.From == null) {
      throw new IllegalStateException("Required field 'From' is missing");
    }
    this.From = builder.From;
    this.To = builder.To == null ? null : Collections.unmodifiableList(builder.To);
    this.CC = builder.CC == null ? null : Collections.unmodifiableList(builder.CC);
    this.BCC = builder.BCC == null ? null : Collections.unmodifiableList(builder.BCC);
    this.Subject = builder.Subject;
    this.Body = builder.Body;
  }
```

2. Nullable annotations for all the fields in builder and non-null/null checks on parameters. This is optional but can increase safety at compile time. IDE flags nullness check redundant.
screen shot 2018-02-05 at 9 06 42 am

### DIFF
--- a/thrifty-java-codegen/src/main/java/com/microsoft/thrifty/gen/ThriftyCodeGenerator.java
+++ b/thrifty-java-codegen/src/main/java/com/microsoft/thrifty/gen/ThriftyCodeGenerator.java
@@ -59,6 +59,7 @@ import com.squareup.javapoet.FieldSpec;
 import com.squareup.javapoet.JavaFile;
 import com.squareup.javapoet.MethodSpec;
 import com.squareup.javapoet.NameAllocator;
+import com.squareup.javapoet.ParameterSpec;
 import com.squareup.javapoet.ParameterizedTypeName;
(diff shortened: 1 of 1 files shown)


benjamin-bader:
Thanks, and nice work!

I'll merge this one, but in the future I'd prefer if we kept unrelated changes in separate PRs - makes reviewing much easier.

PR #167, opened 2018-04-28
Add UrlTransport + Support for strict mode in BinaryProtocol

p-rob:
Mostly based on the official Thrift implementation.

### DIFF
--- a/thrifty-runtime/src/main/java/com/microsoft/thrifty/protocol/BinaryProtocol.java
+++ b/thrifty-runtime/src/main/java/com/microsoft/thrifty/protocol/BinaryProtocol.java
@@ -31,7 +31,7 @@ import java.net.ProtocolException;
 
 /**
  * An implementation of the simple Thrift binary protocol.
- *
+ * <p>
  * <p>Instances of this class are <em>not</em> threadsafe.
  */
 public class BinaryProtocol extends Protocol {
@@ -60,18 +60,49 @@ public class BinaryProtocol extends Protocol {
     private boolean strictRead;
     private boolean strictWrite;
 
-    public BinaryProtocol(Transport transport) {
-        this(transport, -1, -1);
+    public static class Builder {
+
+        private Transport transport;
+        private int stringLengthLimit = -1;
+        private int containerLengthLimit = -1;
+        private boolean strictRead;
+        private boolean strictWrite;
+
+        public Builder(Transport transport) {
+            this.transport = transport;
+        }
+
+        public Builder setStringLengthLimit(int stringLengthLimit) {
+            this.stringLengthLimit = stringLengthLimit;
+            return this;
+        }
+
+        public Builder setContainerLengthLimit(int containerLengthLimit) {
+            this.containerLengthLimit = containerLengthLimit;
+            return this;
+        }
+
+        public Builder setStrictRead(boolean strictRead) {
+            this.strictRead = strictRead;
+            return this;
+        }
+
+        public Builder setStrictWrite(boolean strictWrite) {
+            this.strictWrite = strictWrite;
+            return this;
+        }
+
+        public BinaryProtocol build() {
+            return new BinaryProtocol(this);
+        }
     }
 
-    public BinaryProtocol(Transport transport, int stringLengthLimit) {
-        this(transport, stringLengthLimit, -1);
-    }
-
-    public BinaryProtocol(Transport transport, int stringLengthLimit, int containerLengthLimit) {
-        super(transport);
-        this.stringLengthLimit = stringLengthLimit;
-        this.containerLengthLimit = containerLengthLimit;
+    BinaryProtocol(Builder builder) {
+        super(builder.transport);
+        this.stringLengthLimit = builder.stringLengthLimit;
+        this.containerLengthLimit = builder.containerLengthLimit;
+        this.strictRead = builder.strictRead;
+        this.strictWrite = builder.strictWrite;
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


thrifty-runtime/src/main/java/com/microsoft/thrifty/protocol/BinaryProtocol.java
```diff
@@ -170,8 +201,8 @@ public void writeI16(short i16) throws IOException {
     public void writeI32(int i32) throws IOException {
         buffer[0] = (byte) ((i32 >> 24) & 0xFF);
         buffer[1] = (byte) ((i32 >> 16) & 0xFF);
-        buffer[2] = (byte) ((i32 >>  8) & 0xFF);
-        buffer[3] = (byte)  (i32        & 0xFF);
+        buffer[2] = (byte) ((i32 >> 8) & 0xFF);
```
benjamin-bader: nit: please stick with the pre-existing alignment

thrifty-runtime/src/main/java/com/microsoft/thrifty/protocol/BinaryProtocol.java
```diff
@@ -184,8 +215,8 @@ public void writeI64(long i64) throws IOException {
         buffer[3] = (byte) ((i64 >> 32) & 0xFF);
         buffer[4] = (byte) ((i64 >> 24) & 0xFF);
         buffer[5] = (byte) ((i64 >> 16) & 0xFF);
-        buffer[6] = (byte) ((i64 >>  8) & 0xFF);
-        buffer[7] = (byte)  (i64        & 0xFF);
+        buffer[6] = (byte) ((i64 >> 8) & 0xFF);
```
benjamin-bader: nit: keep the alignment

thrifty-runtime/src/main/java/com/microsoft/thrifty/protocol/BinaryProtocol.java
```diff
@@ -314,31 +345,31 @@ public byte readByte() throws IOException {
     public short readI16() throws IOException {
         readFully(buffer, 2);
         return (short) (((buffer[0] & 0xFF) << 8)
-                       | (buffer[1] & 0xFF));
+                | (buffer[1] & 0xFF));
```
benjamin-bader: nit: alignment.  same for all the other alignment changes in this file.

thrifty-runtime/src/main/java/com/microsoft/thrifty/transport/URLTransport.java
```diff
@@ -0,0 +1,122 @@
...
+    public static class Builder {
+
+        private URL url;
+        private int timeout = 0;
+
+        public Builder(URL url) {
+            this.url = url;
+        }
+
+        public Builder setTimeout(int timeout) {
```
benjamin-bader: 1. What unit is timeout - seconds, millis?  This should be document, and the param name should reflect it, e.g. `timeoutSeconds`.
2. What is this timeout - connect timeout, read timeout, etc?  This should also be documented, and the method name should reflect the purpose.

thrifty-runtime/src/main/java/com/microsoft/thrifty/transport/URLTransport.java
```diff
@@ -0,0 +1,122 @@
...
+
+    @Override
+    public void write(byte[] buffer, int offset, int count) throws IOException {
+        outputStream.write(buffer, offset, count);
+    }
+
+    @Override
+    public void flush() throws IOException {
+
+        HttpURLConnection connection = (HttpURLConnection) url.openConnection();
```
benjamin-bader: I think this logic is in the wrong place.  Evidently, this is connection-establishment and initialization code, which definitely shouldn't be in `flush()`, which can be called any number of times by protocol classes.  Shouldn't this be part of the builder?  Or, at the very least, in an `open()` method of its own?

More worryingly, `inputStream` is being initialized in `flush()`, which is definitely wrong.  That implies that `read()` will always fail before `flush()`.  There's no guarantee that a protocol will `flush()` its transport, so this method is certainly not the place to be preparing for reads.

thrifty-runtime/src/main/java/com/microsoft/thrifty/transport/URLTransport.java
```diff
@@ -0,0 +1,122 @@
...
+import java.net.URL;
+
+public class URLTransport extends Transport {
+
+    private URL url;
+    private int timeout = 0;
+
+    private HttpURLConnection connection;
+    private InputStream inputStream;
+    private ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
```
benjamin-bader: Why not just use the URL connection's OutputStream?

thrifty-runtime/src/main/java/com/microsoft/thrifty/protocol/BinaryProtocol.java
```diff
@@ -31,7 +31,7 @@
 
 /**
  * An implementation of the simple Thrift binary protocol.
- *
+ * <p>
```
benjamin-bader: Did you intend to write more here?  If not, please revert.

thrifty-runtime/src/main/java/com/microsoft/thrifty/transport/URLTransport.java
```diff
@@ -0,0 +1,122 @@
...
+package com.microsoft.thrifty.transport;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.net.HttpURLConnection;
+import java.net.URL;
+
+public class URLTransport extends Transport {
```
benjamin-bader: Please add a doc comment and some unit tests; MockWebServer would be a great tool to use for the latter.

benjamin-bader:
OK, seems clear that this PR is abandoned.  I'll close it accordingly.

PR #173, opened 2018-06-04
Convert thrifty-java-codegen to Kotlin

benjamin-bader:
As part of the prep work for implementing Kotlin codegen, we're converting the compiler to Kotlin.  This is the first portion to be converted; thrifty-schema is next, in stages.


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -2,4 +2,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.3.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.7-all.zip
```
allidoiswarren: 4.8 is out now if you want to use it.

benjamin-bader: Nice!  I'll look into it.

PR #178, opened 2018-07-12
Fix .thrift file matching in include paths

### LINKED ISSUE #177: Bug in loading paths - checking ".thrift" extension on `Path` types
In `Loader`, when loading paths it filters on `.endsWith(".thrift")`. Thing is, it's checking this on `Path` and converting the given parameter to a `Path`. The result is that this filter always fails to find any `.thrift` files:

https://github.com/Microsoft/thrifty/blob/103614bd7e5102a69dc87258d5dee657fa4810ad/thrifty-schema/src/main/java/com/microsoft/thrifty/schema/Loader.java#L194

hzsweers:
Resolves #177


benjamin-bader:
Great catch, and the fix is nice and concise.  Thanks for the contribution!

PR #216, opened 2018-08-08
Fix DoubleValidator support for constants

jparise:
The previous code would still throw IllegalStateException even if a
IdentifierValueElement successfully validated as a correctly-typed
constant.

Fix that by moving the failure case into an `else` clause on the `when`
statement instead of making it the default case for the entire function.

Alternatively, a `return` could be added to the IdentifierValueElement
case, but that's not as idiomatic, and the proposed style matches the
other validators.

This change also adds full test coverage for the DOUBLE validator.


benjamin-bader:
Oh, great catch - thanks for the fix!  Tests look great!

PR #217, opened 2018-08-08
Avoid scope collisions when reading map values

jparise:
Previously, the generated Java code for reading maps would use the same
scope for both the key and the value. If both the key and value were of
the same primitive type (e.g. `map<enum, enum>`), the two intermediate
variable names would clash because they're derived from the same type
and scope (e.g. `i32_1` and `i32_1`).

This change pushes the value reading code into its own nested scope to
avoid that collision (e.g. `i32_1` and `i32_2`).

### DIFF
--- a/thrifty-java-codegen/src/main/kotlin/com/microsoft.thrifty.gen/GenerateReaderVisitor.kt
+++ b/thrifty-java-codegen/src/main/kotlin/com/microsoft.thrifty.gen/GenerateReaderVisitor.kt
@@ -204,9 +204,11 @@ internal open class GenerateReaderVisitor(
             mapType.keyType.accept(this)
             nameStack.pop()
 
-            nameStack.push(value)
-            mapType.valueType.accept(this)
-            nameStack.pop()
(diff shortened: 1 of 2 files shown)


benjamin-bader:
Another good find!

I would have suggested that we could just elide the i32 entirely in `visitEnum`, but I see that we use the value in error messages as well.  I also would have suggested that this scope usage would break for `map<list<enum_t>, enum_t>`, but sure enough it works!  Seems like a tricky corner of Java:

```java
// Error
int x  = 1;
for (...) {
   int x = 3; // illegal redeclaration
}

// But this works
for (...) {
  int x = 3;
}
int x = 1; // No problem
```

Both cases are illegal in C#, at least as far as I remember.  I can see why the second makes sense to allow, but never knew that it was legal Java.  Without a better idea, I'm happy to take this.  Thanks for the contribution!

Incidentally, this is _also_ broken in our Kotlin codegen, but I think we can use `.let` in the `visitEnum` equivalent there to avoid scope tricks.

PR #230, opened 2018-08-17
 Upstreaming rendering tool for thrifty-schema

hzsweers:
This is an upstreaming of our internal renderer for thrifty-schema (and all its element types). This doesn't have any tests included because we don't have any currently, but surfaced up as a starting point :). It does work extremely well though.

High level API overview:
* `ThriftSpec` - represents a thrift spec file. It contains a `Schema` of elements within that spec, includes, namespaces, and the relative file path within the set of specs
* `Include` - represents an `include` statement in a `ThriftSpec`
* `SchemaRendering` - the main entry point for rendering. Can render individual elements or create a multi-file rendered representation of a multi-file `Schema`.

### DIFF
--- a/thrifty-schema/src/main/kotlin/com/microsoft/thrifty/schema/NamespaceScope.kt
+++ b/thrifty-schema/src/main/kotlin/com/microsoft/thrifty/schema/NamespaceScope.kt
@@ -28,7 +28,7 @@ package com.microsoft.thrifty.schema
  * by Thrift and so are included to retain fidelity.  Scopes relevant to
  * Thrifty are [ALL], [JAVA], and [KOTLIN].
  */
-enum class NamespaceScope(private val thriftName: String) {
+enum class NamespaceScope(internal val thriftName: String) {
     /**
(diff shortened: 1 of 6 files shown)


thrifty-schema/src/main/kotlin/com/microsoft/thrifty/schema/NamespaceScope.kt
```diff
@@ -28,7 +28,7 @@ package com.microsoft.thrifty.schema
  * by Thrift and so are included to retain fidelity.  Scopes relevant to
  * Thrifty are [ALL], [JAVA], and [KOTLIN].
  */
-enum class NamespaceScope(private val thriftName: String) {
+enum class NamespaceScope(internal val thriftName: String) {
```
hzsweers: Internally, we'd get this reflectively. Hope making it `internal` is ok!

benjamin-bader:
Thank you for the contribution, and for your patience while I read through it!

PR #232, opened 2018-08-18
Add JVM NamespaceScope option with defaulting

hzsweers:
Should make java/kotlin a bit easier to interop


hzsweers:
The fallback logic and handling is a proposal, happy to revert it. Would ideally want to have something in the annotations that indicates that it's inferred but not sure how to sanely do that on an optional `AnnotationElement`

benjamin-bader:
JVM namespaces and fallbacks are fine; obviously not portable, but anyone who picks it will most likely be aware of the fact.

I suppose if we wanted to note that an annotation value has been inferred, we'd have to explicitly model an `AnnotationValue` type of thing.  Seems like a reasonable thing to do in principle, though I don't understand the use case here .  What use case is enabled by knowing that a value was inferred?

Issue #233, opened 2018-09-04
Consider @Generated annotations


jparise:
The [`@Generated`](https://docs.oracle.com/javase/8/docs/api/javax/annotation/Generated.html) annotation is a signal that a Java source file has been generated programatically. It is used by some static analysis tools, and some code review interfaces automatically "collapse" diffs touching generated files.

It also provides a standard way to include metadata, such as a timestamp. Thrifty currently writes this information as a file comment.

There might be a small cost to including this additional annotation, but I'm not experienced enough in those aspects of today's JVMs to estimate it.

benjamin-bader:
It sounds like a good idea - I wasn't aware that CR tools treated it specially.  Since it is part of the standard JDK (I think?) and has "source" retention, it's not present at all in bytecode and imposes no cost either at runtime or compile-time.

Should be a no-brainer to implement.

benjamin-bader:
Ah, I remember now - we initially did this, then removed the annotations because (absent tool support) they cause tons of noise in code-reviews for teams that check in generated sources.  If we add them back, it should be accompanied by an option to suppress them.

jparise:
Could you expand on that a little bit? Our team checks in generated sources, and we do see some level of noise related to code moving around and the timestamp comment changing. Reducing that noise is my main motivation for proposing `@Generated` because our code review tool (at least) collapses files containing the `@Generated` string.

A compromise might be to just include the literal string `@Generated` in the generated file's header comment, which is what Apache Thrift does.

benjamin-bader:
Sure - per the javadoc, well-formed `@Generated` annotations include a timestamp.  This means that when a .thrift file is updated, _all_ generated code files are touched instead of only those that have meaningful changes.  If code-review tools aren't aware of this annotation, then any review will be dominated by meaningless autogenerated changes.  The simplest solution for us was to just omit the annotation entirely.

We could probably leave the timestamp out, come to think of it - I don't know of any tool that depends on it.  I'd prefer that over going back to comments-as-annotations.

jparise:
That works for me. I can also solve the "auto-collapse in code review" thing locally; `@Generated` just happens to already work.

Are there any other benefits of using the `@Generated` annotation that we haven't covered before we move on? I've been led to understand that some IDEs might treat those files differently (because with regard to refactoring actions?), but I don't have specific examples.

jparise:
[...]

PR #234, opened 2018-09-06
Add an --omit-file-comments command line flag

jparise:
This optional command line flag can be set to disable file comments in
generated (Java) files. The Kotlin code generator doesn't appear to add
any file comments at this time.

See #233


benjamin-bader:
Thank you!

PR #236, opened 2018-09-11
Only use base filename in constants' comments

jparise:
This line unfortunately included the local file system path that can
differ between individual systems. This isn't ideal for projects that
store these generated files in source control because is causes
unnecessary file differences.

See #233


benjamin-bader:
I'd like to keep some form of this comment; fields in the same `Constants` class can come from different .thrift files, and in that case it's useful to know their sources.  Printing only some kind of relative path would eliminate unnecessary file diffs, but still keep this information.

jparise:
Do you think stripping the path down to just its `filename.thrift` component would be unique enough?

benjamin-bader:
I'm thinking we could probably use `Location#path`, which is (IIRC) the portion of the path relative to the compiler's working directory.  Would need to verify that, as I haven't looked at that part of the code recently.

benjamin-bader:
hmm, my memory is rusty, and location handling seems to be kind of busted as-is - as far as I can tell, `path` only ever seems to contain the actual file name, e.g. `filename.thrift`.  It _should_ have the relative include path, but that's another bug.

I'd say in the meantime, let's just use `path` as-is since that's already an improvement, and I'll file a bug to look at location handling in the future.

PR #241, opened 2018-09-14
Add @Nullable annotation to findByValue methods

jparise:
The default clause returns null so these methods should be marked as
nullable when Android annotations are enabled.


jparise:
@NightlyNexus I went with an even simpler approach. Many of the other tests construct their own `ThriftyCodeGenerator` directly, and that works well here.

NightlyNexus:
Nice!

benjamin-bader:
Thanks, folks!

PR #242, opened 2018-09-25
Update to Okio 2.1.0.

NightlyNexus:
Adds the Kotlin stdlib dependency. dunno what your thoughts on migration there is.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -56,7 +56,7 @@ allprojects {
                 ],
 
                 okio: [
-                        'com.squareup.okio:okio:1.14.1'
+                        'com.squareup.okio:okio:2.1.0'
                 ],
 
                 testing: [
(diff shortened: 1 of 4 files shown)


benjamin-bader:
Looks like integration tests are broken by this PR; should be a simple fix, by the looks of it.

I'm a little leery of this change as-is.  As you mention, this will add the kotlin-stdlib dependency to `thrifty-runtime`, which I've been avoiding in favor of providing a `thrifty-runtime-ktx` module.  If okio weren't such a fundamental library these days, I'd reject the change out-of-hand.

What do you think about separating the compile-time and runtime okio versions - is that too crazy?

benjamin-bader:
Ping - is there still interest in this change on your end?

benjamin-bader:
Closing this due to lack of response and drift in the master branch.  If there's any more interest in moving Thrifty to Okio 2, please open an issue so we can work through the dependency implications.

Given the upcoming bump to 2.0.0, we have a nice window to make bigger changes like this.

PR #243, opened 2018-09-30
Replace compile with implementation

rosariopfernandes:
Most recent versions of gradle now use `compile` instead of `implementation`


benjamin-bader:
Good point - thanks for the fix!

PR #245, opened 2018-10-04
Update CONTRIBUTING.md

doilio:


benjamin-bader:
This is an interesting corner of grammar!  As a native speaker of English I don't really think about grammar, and couldn't say why this proposed change felt wrong.  A little bit of research turned up an interesting article: https://www.quickanddirtytips.com/education/grammar/dos-and-donts

It seems that there are diverging opinions here.  Personally, I'm inclined to say that the apostrophe in "don't" is because the word is a contraction of "do not", and not because the word is plural.  I'm going to leave the text as-is.

PR #248, opened 2018-10-05
Chanded the gradle distribution Url

RichaldoElias:


benjamin-bader:
I have no idea what the intent is here and haven't heard back, so I'm going to both of these PRs.

Issue #244, opened 2018-10-02
Constructing Builders with nullable structs


jparise:
As I convert more and more (Java) code from Apache Thrift to Thrifty, I've started to run into a pattern where I start with an existing struct object that _might_ be `null` and then "extends" that struct by setting some existing values. The most direct way to do this is by conditionally calling the correct Builder construct:

```java
public Struct updateStruct(Struct struct) {
    Struct.Builder builder = (struct == null)
        ? new Struct.Builder()
        : new Struct.Builder(struct);

    builder.foo(True);

    return builder.build();
}
```

It would be much more convenient if `Struct.Builder` accepted a nullable argument and added a `if (struct != null)` guard to the generated code so the above could just become:

```java
public Struct updateStruct(Struct struct) {
    return new Struct.Builder(struct)
        .foo(True)
        .build();
}
```

That would detract a bit from the purity of the existing Builder interface, but this could be a case where caller convenience could take precedence.

jparise:
I'll also acknowledge that some larger refactoring exercises are in order, such as passing Builder objects through this calls instead of reified structs. Please take the above primarily in terms of an initial conversion and less so as a critique of the current API.

benjamin-bader:
The need for this certainly does feel like a sign that some refactoring is needed, but in any event I'm not opposed to convenience here.

jparise:
As an incremental improvement, #250 adds an explicit `@NonNull` annotation.

PR #250, opened 2018-10-05
Add @NonNull to struct builder's copy constructor

jparise:
The builder's copy constructor receives a `struct` value that can't be
null. Add `@NonNull` when Android annotations are enabled.

See #244


benjamin-bader:
Looks good to me - thanks!

PR #252, opened 2018-10-07
Add newer implementation of gradle-wrapper

beatbrot:
The gradlewrapper.jar is rather old (4.3.1).

I updated it to a newer version (4.10.2). Even tough the wrapper is newer, the used gradle-version for building is determined by `gradle.properties`

This shouldn't add new problems, I suppose


benjamin-bader:
> the used gradle-version for building is determined by gradle.properties

I'm not sure I understand - are you saying that the `gradle.properties` file at the root of the thrifty repo determines which Gradle version is used to compile the project?

beatbrot:
> 
> 
> > the used gradle-version for building is determined by gradle.properties
> 
> I'm not sure I understand - are you saying that the `gradle.properties` file at the root of the thrifty repo determines which Gradle version is used to compile the project?

Sorry, actually I meant the [gradle-wrapper-properties](https://github.com/Microsoft/thrifty/blob/master/gradle/wrapper/gradle-wrapper.properties) file

benjamin-bader:
Thanks, that makes much more sense.  I see now that this PR only includes the jarfile itself - I think that the wrapper properties should be updated in tandem with the jar.  Would you be willing to add that here?

beatbrot:
Actually that is exactly what I did: I called the wrapper task with my local Gradle installation (4.10.2) and committed the newly generated wrapper

benjamin-bader:
To be more clear, I was referring to adding task settings (version number, distribution type, etc) in the root build.gradle file:
```gradle
wrapper {
    gradleVersion = "4.10.2"
    distributionType = Wrapper.DistributionType.ALL
}
```

Even better than simply upgrading the wrapper, this would let us know just by looking what version of the wrapper we're using.

beatbrot:
That makes a lot of sense! 
Since you actually provided the whole solution and I don't want to piggy-bank off of you:
 Do you want to create your own issue or should I just update mine? :)

benjamin-bader:
Almost there - you've set the version in your code to 4.7, but the wrapper jar itself is 4.10.2.  Which version did you intend to check in?

beatbrot:
4.7 is there for a reason!

Apparently the option specified in the wrapper-task do only determine the distributionUrl in gradle-wrapper.properties. They don't determine which version of the wrapper itself will be used.

So what I want is, that the latest wrapper (4.10.2) should download Gradle version 4.7.

Therefore I set gradleVersion to 4.7!

Issue #185, opened 2018-07-30
Parser should allow binary literals


benjamin-bader:
They aren't explicitly supported by the Apache compiler, but I don't think they're _disallowed_, either.  They'd be neat.  Let's suppose that they should be hex strings.

The change would need to be in the ANTLR grammar first, then in the validation logic in `Constant.kt`.

benjamin-bader:
This feature would allow the following Thrift syntax:

```thrift
struct Example {
  1: binary SomeBinary = "0xDEADBEEF"
}
```

picimaci:
I don't have any experience with ANTLR grammar yet but I would gladly give this a go if it's okay :)

benjamin-bader:
Please, go ahead!

PR #256, opened 2018-10-31
Add binary literals to parser

picimaci:
Added binary literals to grammar and code. Did it best to my knowledge however I probably got things wrong at places. Thank you in advance for reviewing :)
Will solve #185 .

### DIFF
--- a/thrifty-schema/src/main/kotlin/com/microsoft/thrifty/schema/Constant.kt
+++ b/thrifty-schema/src/main/kotlin/com/microsoft/thrifty/schema/Constant.kt
@@ -234,8 +234,14 @@ class Constant private constructor (
     private object BinaryValidator : BaseValidator() {
         override fun validate(symbolTable: SymbolTable, expected: ThriftType, valueElement: ConstValueElement) {
             when (valueElement) {
-                is BinaryValueElement -> return
-                else -> super.validate(symbolTable, expected, valueElement)
+                is LiteralValueElement -> {
+                    val text = valueElement.value
+                    val hexRegex = Regex("0[Xx][a-fA-F0-9]+")
+
+                    if(!hexRegex.matches(text)) {
+                        throw IllegalStateException("Expected binary value as hex but got $text")
+                    }
+                }
             }
         }
     }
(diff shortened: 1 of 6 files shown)


benjamin-bader:
Hi, thanks for the PR, I'm looking forward to reviewing it!  Looks like the unit tests failed in a way that's unrelated to your change, so I restarted them.  It's a busy time right now, so thanks for your patience w/review.

benjamin-bader:
The ANTLR looks perfectly proficient, though I wonder about the definition of the BINARY token.  As defined, it overlaps with STRING and consequently it looks to me like it might either never be matched, or might break some valid string constants.  I'd like to see some unit tests in `LoaderTest` that exercise the whole loader:

```thrift
// these would each be separate test cases
const string STRING_WITH_BINARY_LITERAL = "0xDEADBEEF" // should load

const binary VALID_BINARY = "0xDEADBEEF" // should load

const binary INVALID_BINARY = "blablabla" // should not load
```

I might be wrong here.  If what I suspect is true, the solution will be to drop the changes to the ANTLR grammar, and simply use a STRING token; we'd have to validate its format by hand, in the BinaryValidator.

picimaci:
Or by the next couple of days, sorry for the delay.

benjamin-bader:
Hey, don't feel pressured - there's no rush.  Take your time!

picimaci:
Hey! Any feedback? :)

benjamin-bader:
Sorry, got busy with travel over the past few weeks, and haven't had the time yet to look at the update.  I'll take a look now.

thrifty-schema/src/main/kotlin/com/microsoft/thrifty/schema/Constant.kt
```diff
@@ -239,6 +231,21 @@ class Constant private constructor (
         }
     }
 
+    private object BinaryValidator : BaseValidator() {
+        override fun validate(symbolTable: SymbolTable, expected: ThriftType, valueElement: ConstValueElement) {
+            when (valueElement) {
```
benjamin-bader: This `when` needs an `else -> super.validate(symbolTable, expected, valueElement)`.  Without it, we won't properly validate identifiers of binary constants.  We should have a test case for that, too:

```thrift
const binary FOO = "0xDEADBEEF"

struct S {
  1: required binary b = FOO;
}
```

picimaci: Totally forgot about this, I'll fix it during the holidays.

benjamin-bader:
Hi there - this PR is nearly ready to land.  Do you still have interest in finishing it?

benjamin-bader:
Hello again.  Any interest in finishing this?  I'll close it out if not.

PR #259, opened 2018-11-02
Support configurable @Generated annotation types

### LINKED ISSUE #257: Fix @Generated JDK9 mess
`javax.annotation.Generated` no longer exists (or is merely inaccessible?  can't tell yet) in JDK9.  Instead, there is the more-specific `javax.annotation.processor.Generated`, which is maybe not what we want (as it is specific to annotation processors).

This is currently a problem because folks trying to build Thrifty using Java 9+ hit a few associated compiler errors.  It's not sufficient to remove direct references to the `@Generated` annotation in the compiler, because the integration tests would still fail to compile.

A few things need to happen here:
1. Remove direct references to `@Generated`
2. Make option to specify none, old, or new annotation, both for Java and Kotlin.
3. Add command-line option for 2.

benjamin-bader:
Fixes #257


README.md
```diff
@@ -481,7 +481,13 @@ The final new flag is `--kt-file-per-type`.  Thrifty's convention is to generate
 
 `--omit-file-comments` suppresses the comment that prefixes each autogenerated Java file.
 
-Types generated by Thrifty are annotated with `@javax.annotation.Generated` by default; this can be disabled with the `--omit-generated-annotations` flag.
+Types generated by Thrifty are annotated with a `@Generated` annotation by default; this can be disabled with the `--omit-generated-annotations` flag.  The actual type of the annotation may be specified with a `--use-generated-annotation=[jdk8|jdk9|native]` argument.  By default, we assume `jdk8`, i.e. `javax.annotation.Generated`, as this is what's supported in the Android toolchain ()and will be for the foreseeable future).
```
jparise: I think the `()` here is meant to be just `(`.

benjamin-bader: Yep!  And now the typo is enshrined in v1.0.0 forever :)

PR #273, opened 2018-11-27
Improve Program Circular Include Message

muandrew:
Adding more info in the exception to make identifying the circular
include more easily.


benjamin-bader:
Thanks for contributing!

PR #282, opened 2018-12-20
Strip trailing spaces in rendering and ignore in equality checking

hzsweers:
Saves a few bytes. Specifically would happen on cases where docs might have trailing spaces and avoids doing unnecessary trailing spaces in blank lines between docs


thrifty-schema/src/main/kotlin/com/microsoft/thrifty/schema/SchemaFunctionalEquality.kt
```diff
@@ -32,7 +32,7 @@ package com.microsoft.thrifty.schema
  * For comparing docs, we remove stars as things get weird in the parser.
  */
 private fun String.cleanedDoc(): String {
-    return trim().replace("*", "")
+    return trim().replace("*", "").split("\n").joinToString("\n") { it.trimEnd() }
```
benjamin-bader: Nice!

PR #284, opened 2019-01-07
Expose read functionality from Adapter through Struct interface

vladimirpetrovski:
I've made this pull request as it is a small change and might be a good change. Code generation was fixed for both languages (Java and Kotlin) as well.

Similarly as `Adapter#write()` was exposed, I'm exposing it `Adapter#read()` as well.

This is a small change and not necessarily a valid case. 

Looking forward to your feedback.

Thank you!

### DIFF
--- a/thrifty-java-codegen/src/main/kotlin/com/microsoft.thrifty.gen/ThriftyCodeGenerator.kt
+++ b/thrifty-java-codegen/src/main/kotlin/com/microsoft.thrifty.gen/ThriftyCodeGenerator.kt
@@ -222,7 +222,7 @@ class ThriftyCodeGenerator {
 
         val structBuilder = TypeSpec.classBuilder(type.name)
                 .addModifiers(Modifier.PUBLIC, Modifier.FINAL)
-                .addSuperinterface(Struct::class.java)
+                .addSuperinterface(ParameterizedTypeName.get(ClassName.get(Struct::class.java), structTypeName))
 
         if (type.hasJavadoc) {
             structBuilder.addJavadoc("\$L", type.documentation)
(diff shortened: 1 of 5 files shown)


benjamin-bader:
Hi there, thanks for contributing - it's not a small thing to just jump in to the code generators!

This would be an API-breaking change for consumers (I think), which we strongly prefer to avoid unless there are clear benefits.  That said, I'm not sure what the use-case is here.  Can you please explain what situations this change helps with, and how?

vladimirpetrovski:
Hi Benjamin! Thanks for taking a look at the pull request. 

Currently, there is one problem I can remember of in my case, that can be solved by this pull request.

Let's say we have a serialized thrift object, that I want to deserialize it into a certain struct object (which is done by the adapter in each `Struct` implementation). The problem here is that there is no generic way to access that `adapter.read()` function in order to deserialize the object. In my case, I receive a byte array and trying to use a `BinaryProtocol`, in order to deserialize it into certain `Struct` implementation.

For example:
```
  public <T extends Struct> T deserialize(T message, byte[] bytes) throws IOException {
    Buffer buffer = new Buffer();
    BinaryProtocol binaryProtocol = new BinaryProtocol(new BufferTransport(buffer));
    buffer.write(bytes);
    return message.read(binaryProtocol);
  }
```

Does it make sense? Or can you give me another solution for this kind of problem?

Thanks!

benjamin-bader:
Ah, I see.  We've done this using only adapters.  The only trouble is in denoting a generic type that extends Adapter, which has a complicated set of parameters, but you can do it like so:

```java
public <S extends Struct, B extends StructBuilder<S>, A extends Adapter<S, B>> deserialize(byte[] bytes, A adapter) throws IOException {
  BinaryProtocol proto = protocolFromBytes(bytes);
  return adapter.read(proto);
}

// Use like so:
MyStruct s = deserialize(bytes, MyStruct.ADAPTER); // possibly you'd have to specify the generics, I don't remember
```

Admittedly ugly, but it works.

benjamin-bader:
Does my suggestion above work for you?  Please let me know, so we can close this issue (assuming it works).

benjamin-bader:
Since I haven't heard back in three weeks, I'm going to assume that your needs are now met and close the PR.  Please re-open if that's not the case.  Thanks!

PR #305, opened 2019-02-13
Specify where a missing file was included in failure message

ZacSweers:
This makes debugging easier since looking at that specific file is easier than trying to deduce which one used that namespace/directory


benjamin-bader:
Looks great, thank you!

PR #319, opened 2019-08-21
Travis: openjdk8, openjdk11

sullis:


benjamin-bader:
Thanks, this is a good move, especially as JDK10 is end-of-life.

I think you'll need to remove the entire `addons` section from the end of the config file as well - looks like _something_ related to installing Oracle JDK8 is failing, and that's the only thing I can see that looks related.

benjamin-bader:
Fixed in #321.

PR #330, opened 2020-05-07
Introduce an --omit-service-clients option

jparise:
When specified, no service clients will be generated. We only support
service clients in the Kotlin generator, so this is implicitly a
Kotlin-only option, but it's not prefixed with `kt` because it could
apply to Java in the future.

This is useful for projects that only need Thrift-generated structures,
etc. and won't use Thrift service calls.


benjamin-bader:
Looks great - thanks!

PR #331, opened 2020-05-12
Remove unneeded/temporary code lines

jparise:
It doesn't look like these lines of code were intended to be committed.


benjamin-bader:
lol

PR #332, opened 2020-05-12
Introduce a --kt-emit-jvmname option

jparise:
When specified, `@file:JvmName(name)` annotations will be added to all
of the generated files. This improves the usability of importing these
Kotlin-based classes into Java code modules (e.g. as `foo.Constants`
instead of `foo.ConstantsKt`).

This would ideally be the default behavior, but it's introduced as an
opt-in compiler option to avoid breaking backwards compatibility.

### DIFF
--- a/thrifty-kotlin-codegen/src/main/kotlin/com/microsoft/thrifty/kgen/KotlinCodeGenerator.kt
+++ b/thrifty-kotlin-codegen/src/main/kotlin/com/microsoft/thrifty/kgen/KotlinCodeGenerator.kt
@@ -65,6 +65,7 @@ import com.microsoft.thrifty.service.TMessageType
 import com.microsoft.thrifty.util.ObfuscationUtil
 import com.microsoft.thrifty.util.ProtocolUtil
 import com.squareup.kotlinpoet.AnnotationSpec
+import com.squareup.kotlinpoet.AnnotationSpec.UseSiteTarget.FILE
 import com.squareup.kotlinpoet.BOOLEAN
 import com.squareup.kotlinpoet.ClassName
 import com.squareup.kotlinpoet.CodeBlock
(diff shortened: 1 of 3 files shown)


thrifty-kotlin-codegen/src/main/kotlin/com/microsoft/thrifty/kgen/KotlinCodeGenerator.kt
```diff
@@ -2219,6 +2225,17 @@ class KotlinCodeGenerator(
         return block.build()
     }
 
+    private fun makeFileSpecBuilder(packageName: String, fileName: String): FileSpec.Builder {
```
jparise, line 2228: I started with a `FileSpec.Builder.addJvmName` extension function, but once I made this optional behavior, I thought it made more sense to wrap the entire `FileSpec.Builder` creation pattern in a single function that either added the `JvmName` annotation or didn't.

Happy to use a different pattern based on feedback, though!

thrifty-kotlin-codegen/src/test/kotlin/com/microsoft/thrifty/kgen/KotlinCodeGeneratorTest.kt
```diff
@@ -365,6 +365,62 @@ class KotlinCodeGeneratorTest {
...
+        """.trimMargin()
+
+        val text = generate(thrift) {
+                    emitJvmName()
+                    filePerNamespace()
+                }
+                .single()
+                .toString()
+
+        text shouldBe """
```
jparise, line 383: These tests could be a little simpler with the `shouldStartWith` matcher (because we're only interested in the first line of output), but that doesn't seem to be available in this version of kotlintest (kotest).

thrifty-kotlin-codegen/src/main/kotlin/com/microsoft/thrifty/kgen/KotlinCodeGenerator.kt
benjamin-bader: This seems fine to me!

benjamin-bader:
Looks fine to me, thank you!  We could definitely stand to update our testing deps, but I won't block this PR on that.

PR #353, opened 2020-05-17
Apply checkstyle only to thrifty-runtime

benjamin-bader:
This is our last and only Java module.  There's no point to applying checkstyle to any other module in our project - it only burns CPU and causes problems.


benjamin-bader:
OK I was wrong on "last java module".  The true statement is "last _important_ java module".

Issue #358, opened 2020-06-02
A client request never completes when an non-expected exception occurs 


amorozov:
If an exception occurs when performing a client request (i.e. an IOException or a RuntimeException) then the request never completes, even with a fail.

This happens because these types of exceptions are handled differently in `AsyncClientBase.invokeRequest()` than other types, and the exception is re-thrown without `fail`-ing the request leaving client code hung on the request.

A simple patch can fix the problem:

```
diff --git a/thrifty-runtime/src/main/java/com/microsoft/thrifty/service/AsyncClientBase.java b/thrifty-runtime/src/main/java/com/microsoft/thrifty/service/AsyncClientBase.java
index c944b67..288a08e 100644
--- a/thrifty-runtime/src/main/java/com/microsoft/thrifty/service/AsyncClientBase.java
+++ b/thrifty-runtime/src/main/java/com/microsoft/thrifty/service/AsyncClientBase.java
@@ -203,6 +203,7 @@ public class AsyncClientBase extends ClientBase implements Closeable {
             try {
                 result = AsyncClientBase.this.invokeRequest(call);
             } catch (IOException | RuntimeException e) {
+                fail(call, e);
                 throw e;
             } catch (ServerException e) {
                 error = e.thriftException;
```

benjamin-bader:
Good catch - happy to take a PR with this patch!

benjamin-bader:
Fixed in #362 - thanks again!

PR #362, opened 2020-06-05
Explicitly call fail() on a failed request when an exception occurs

amorozov:
If an exception occurs when performing a client request (i.e. an
IOException or a RuntimeException) then the request never completes,
even with a fail.

This happens because these types of exceptions are handled differently
in `AsyncClientBase.invokeRequest()` than other types, and the
exception is re-thrown without `fail`-ing the request leaving
client code hung.

This patch fixes the problem #358.

Signed-off-by: Alexey Morozov <morozov@gmail.com>


benjamin-bader:
Thanks for the patch!

PR #367, opened 2020-06-16
Add API to configure which specific thrift files get compiled

### LINKED ISSUE #365: Add sourceFiles to thrifty-gradle-plugin
The Gradle plugin currently supports parsing directories of `.thrift` source files via `sourceDir` / `sourceDirs`. In our environment, we only want to select specific `.thrift` files out of a source directory.

I don't know Gradle enough to estimate how easy it would be to add something like a `sourceFiles` directive:

```gradle
sourceFiles 'path/to/file.thrift', 'path/to/another/file.thrift'
```

... but that would be a good fit for our use case.

benjamin-bader:
Adds an API to allow configuring include and exclude patterns on `sourceDir` entries in the Gradle plugin extension.

@jparise please take a look and let me know whether this addresses your use-case.  You're welcome to comment on anything here but probably the most interesting thing will be the readme changes which document the new API, and its direct implementation in `ThriftyExtension.kt` should you be so inclined.

Fixes #365

### DIFF
--- a/thrifty-gradle-plugin/README.md
+++ b/thrifty-gradle-plugin/README.md
@@ -17,9 +17,24 @@ thrifty {
     // By default, thrifty will use all thrift files in 'src/main/thrift'.
     // You can override this in the following ways:
 
-    // Specify one or more directories containing thrift tiles.
+    // Specify a directory containing thrift files.
     // Note that if you do, 'src/main/thrift' is ignored.
+    // By default, all .thrift files in this directory and its subdirectores
+    // will be compiled.
     sourceDir 'path/to/thrift/files', 'path/to/other/thrift/files'
+
+    // Specify several directories containing thrift files, as above.
+    // Note the plural name 'sourceDirs'.
+    sourceDirs 'path/to/thrift/files', 'path/to/other/thrift/files'
+
+    // Specify a directory containing thrift files along with glob patterns
+    // identifying the exact files to include or exclude.
+    sourceDir('path/to/thrift/files') {
+        // Include and exclude patterns as per:
+        // https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/util/PatternFilterable.html
+        include '**/*.thrift'
+        exclude '**/*.analytics.thrift'
+    }
 
     // Augment the thrift include path with one or more directories:
     includeDir 'path/to/include/dir', 'path/to/other/dir'
--- a/thrifty-gradle-plugin/src/main/kotlin/com/microsoft/thrifty/gradle/ThriftyExtension.kt
+++ b/thrifty-gradle-plugin/src/main/kotlin/com/microsoft/thrifty/gradle/ThriftyExtension.kt
@@ -21,10 +21,12 @@
 package com.microsoft.thrifty.gradle
 
 import org.gradle.api.Action
+import org.gradle.api.file.SourceDirectorySet
 import org.gradle.api.model.ObjectFactory
 import org.gradle.api.provider.ListProperty
 import org.gradle.api.provider.Property
 import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.InputFiles
 import org.gradle.api.tasks.Optional
 import java.util.SortedMap
 import java.util.TreeMap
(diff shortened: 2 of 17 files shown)


thrifty-gradle-plugin/README.md
```diff
@@ -17,10 +17,25 @@ thrifty {
...
+    // identifying the exact files to include or exclude.
+    sourceDir('path/to/thrift/files') {
+        // Include and exclude patterns as per:
+        // https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/util/PatternFilterable.html
+        include '**/*.thrift'
+        exclude '**/*.analytics.thrift'
+    }
+
     // Augment the thrift include path with one or more directories:
     includeDir 'path/to/include/dir', 'path/to/other/dir'
```
jparise, line 40: Unrelated, but this should be `includePath`.

jparise:
I'm try to get this running locally so I can run through it fully, but I could use a little help that may be related to how I'm building/using this branch locally.

I'm able to load the plugin with a little `buildscript` help, but something as simple as this:

```gradle
buildscript {
    repositories {
        flatDir { dirs /* paths to locally-built jars */ }
    }
    dependencies {
        classpath 'com.microsoft.thrifty:thrifty-gradle-plugin:2.1.1-SNAPSHOT'
    }
}

apply plugin: 'com.microsoft.thrifty'
apply plugin: 'kotlin'

thrifty {
    kotlin {
        omitFileComments true
    }
}
```

... barks back at me this error when I run the `generateThriftFiles` task:

```
A problem occurred evaluating project ':modules:schemas'.
> No signature of method: schemas_lfo0wux2ge0pedca69lbazbc.thrifty() is applicable for argument types: (schemas_lfo0wux2ge0pedca69lbazbc$_run_closure1) values: [schemas_lfo0wux2ge0pedca69lbazbc$_run_closure1@707e1a69]
  Possible solutions: notify(), print(java.lang.Object), print(java.io.PrintWriter), print(java.lang.Object)
```

`thrifty { kotlin {} }` (which I see through the test suite) "works" because it doesn't have any closures, but that's not too helpful in my case.

benjamin-bader:
Yikes, that sure is unfriendly.  The error is that `omitFileContents` isn't a valid option.  I'll see if there's a way to make that less cryptic.

jparise:
> Yikes, that sure is unfriendly. The error is that `omitFileContents` isn't a valid option. I'll see if there's a way to make that less cryptic.

I don't think that's quite it in my case. I run into the same problem when I use a valid option (thanks for spotting my goof!) or something like:

```groovy
thrifty {
    includePath 'commons/thrift'
    kotlin { }
}
```

benjamin-bader:
Again, I'm sorry to say that your example isn't well formed.  As you noted, while the option _should_ be `includePath` it is in fact `includeDir`.

jparise:
Ah. And correcting that gets me to an error like this:

```
> Failed to query the value of task ':modules:schemas:generateThriftFiles' property 'includePath'.
   > Cannot convert the provided notation to an object of type Dependency: /<removed>/modules/schemas/thrift.
     The following types/formats are supported:
       - Instances of Dependency.
       - String or CharSequence values, for example 'org.gradle:gradle-core:1.0'.
       - Maps, for example [group: 'org.gradle', name: 'gradle-core', version: '1.0'].
       - FileCollections, for example files('some.jar', 'someOther.jar').
       - Projects, for example project(':some:project:path').
       - ClassPathNotation, for example gradleApi().
```

... for `includePath 'thrift'`. If I pass a non-existent directory name instead, it gives me a error, so it appears to be finding the path just fine.

thrifty-gradle-plugin/README.md
jparise: Can `includeDir` receive multiple paths, or would it need to be `includeDirs` in this example?

benjamin-bader:
That one is clearly down to my misunderstanding of the dependency APIs; will fix.

jparise:
I'm afraid cfa1483 didn't fix that error for me.

jparise:
Also, if I use something like:

```gradle
thrifty {
    sourceDir 'thrift'
}
```

... I get `path must be a directory`. If I use a non-existent directory name, I don't receive any error.

And if I try:

```gradle
thrifty {
    sourceDir('thrift') {
        include 'base.thrift'
    }
}
```

... I get this unfriendly result:

```
> No signature of method: schemas_lfo0wux2ge0pedca69lbazbc.thrifty() is applicable for argument types: (schemas_lfo0wux2ge0pedca69lbazbc$_run_closure1) values: [schemas_lfo0wux2ge0pedca69lbazbc$_run_closure1@95e3289]
```

benjamin-bader:
There must be something very different between the integration tests and your actual usage, because I can't seem to reproduce most of the issues you're encountering :(

thrifty-gradle-plugin/README.md
benjamin-bader: The rename seems reasonable.

`includeDir` takes one single path; `includeDirs` is the varargs version.

benjamin-bader:
Thanks for spending so much time with this today.  I finally got a repro, but have to confess that I don't quite understand _why_ the error happens.

As with most things in Gradle, the problem goes away when I stop doing things in the officially-recommended way.

jparise:
Now I'm stuck trying to resolve this (apparently common) ANTLR dependency issue that pops up sometimes in Android projects:

```ANTLR Tool version 4.8 used for code generation does not match the current runtime version 4.5.3ANTLR Runtime version 4.8 used for parser compilation does not match the current runtime version 4.5.3```

benjamin-bader:
Lol, wtf.  Could it be that you are depending on multiple thrifty versions (one for the plugin, another for the rest of it)?

jparise:
I think aside from the fact that I can't quite test it in-to-end, the code in this branch should do what I want it to. :)

The one additional thing that would be nice to customize is the output directory. There doesn't appear to be an option for that.

benjamin-bader:
Can you say more about the output directory?  I'm thinking that the files generated here are ephemeral, not to be preserved or checked in.  What would configuring the output directory enable for you?

jparise:
> Can you say more about the output directory? I'm thinking that the files generated here are ephemeral, not to be preserved or checked in. What would configuring the output directory enable for you?

It's not a very important requirement. We currently generate the files in `$projectDir/src/main/kotlin` and check them into source control (for historic reasons so that every build didn't need to have the `thrift` binary back in our pre-Thrifty days). We can revisit this, but for the sake of parity, I was going to configure the output path to be the same as it is currently.

PR #368, opened 2020-06-26
Fixed infinity loop in FramedTransport.read in case of EOF.

denisov-mikhail:
FramedTransport.readHeader() doesn't handle case when underlying socket stream returns -1 (EOF), it is lead to infinity loop.

The problem is stable reproduced when server side is crashed.

### DIFF
--- a/thrifty-runtime/src/main/java/com/microsoft/thrifty/transport/FramedTransport.java
+++ b/thrifty-runtime/src/main/java/com/microsoft/thrifty/transport/FramedTransport.java
@@ -23,6 +23,7 @@ package com.microsoft.thrifty.transport;
 import com.microsoft.thrifty.util.UnsafeByteArrayOutputStream;
 
 import java.io.IOException;
+import java.io.EOFException;
 
 /**
  * A transport decorator that reads from and writes to the underlying transport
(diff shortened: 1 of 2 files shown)


thrifty-runtime/src/main/java/com/microsoft/thrifty/transport/FramedTransport.java
```diff
@@ -62,7 +62,9 @@ public int read(byte[] buffer, int offset, int count) throws IOException {
 
     private void readHeader() throws IOException {
         byte[] headerBytes = new byte[4];
-        inner.read(headerBytes, 0, headerBytes.length);
+        int numRead = inner.read(headerBytes, 0, headerBytes.length);
```
benjamin-bader, line 65: As long as we're correcting this oversight, let's make sure we handle partial reads:

```
int numRead = 0;
while (numRead < headerBytes.length) {
  int n = inner.read(headerBytes, numRead, headerBytes.length - numRead);
  if (n == -1) {
    throw new EOFException();
  }
  numRead += n;
}
```

denisov-mikhail: Hi, thanks for review ) It's done! Maybe I wrong, but I think that two consecutive partial reads never happens because here data is reading in blocking way i.e. current thread just blocking until all needed data are read. If partial read is happen that mean no data left in stream and EOF is follow by.

benjamin-bader: The thing here is that `Transport#read` makes no guarantee about that, so we have to assume that `read` can return before fully reading everything requested.  It's possible that `Buffer` makes that guarantee - I don't remember offhand - but certainly socket input streams can return without reading everything requested.

A partial read means EOF, certainly - but here it also means an incomplete or malformed data stream, possibly corrupted data.  Should we close un-exceptionally and just return `-1`?  I don't think it's clear that we should.  What do you think?

denisov-mikhail: Sorry for tardy answer. I think if we here read not enough data to parse header the best way is throw an exception because here is protocol violation, but -1 is some kind of 'normal' behavior.

benjamin-bader:
Looks great - thanks for your contribution!

PR #373, opened 2020-07-15
Try to force the correct ANTLR version in the gradle plugin

benjamin-bader:
@jparise please try 2.1.2-SNAPSHOT once this lands and see if you still have the same bad behavior

### DIFF
--- a/thrifty-gradle-plugin/build.gradle
+++ b/thrifty-gradle-plugin/build.gradle
@@ -46,6 +46,7 @@ gradlePlugin {
 dependencies {
     compileOnly gradleApi()
 
+    api libraries.antlr
     api libraries.kotlin
 
     implementation project(':thrifty-java-codegen')


jparise:
This gets me a bit further, and it's possible what I'm now seeing is due to the way I'm mutating the buildscript's classpath to find the locally-build snapshot, but I now get this:

```
Caused by: java.lang.NoClassDefFoundError: org/antlr/v4/runtime/CharStreams
        at com.microsoft.thrifty.schema.parser.ThriftParser.parse(ThriftParser.kt:52)
        at com.microsoft.thrifty.schema.Loader.loadSingleFile(Loader.kt:240)
        at com.microsoft.thrifty.schema.Loader.loadFileRecursively(Loader.kt:195)
        at com.microsoft.thrifty.schema.Loader.loadFileRecursively$default(Loader.kt:183)
        at com.microsoft.thrifty.schema.Loader.loadFromDisk(Loader.kt:149)
        at com.microsoft.thrifty.schema.Loader.load(Loader.kt:120)
        at com.microsoft.thrifty.gradle.ThriftyTask.run(ThriftyTask.kt:72)
```

(`org/antlr/v4/runtime/CharStreams` is the symbol that doesn't exist in ANTLR 4.5.3; it was introduced in 4.7, I think.)

If I "force" the ANTLR dependency (`classpath 'org.antlr:antlr4:4.8-1'`), I end up with the old error message:

```
ANTLR Tool version 4.8 used for code generation does not match the current runtime version 4.5.3ANTLR Runtime version 4.8 used for parser compilation does not match the current runtime version 4.5.3
```

PR #391, opened 2020-08-24
Convert thrifty-runtime to kotlin, require Java 8

### LINKED ISSUE #387: Move to 100% Kotlin
With such fundamental libraries as okio bringing in Kotlin dependencies, we're no longer doing anyone any method-count favors at this point by remaining Java-only in the runtime.  We should combine `thrifty-runtime` and `thrifty-runtime-ktx` and rewrite in Kotlin.

### LINKED ISSUE #388: Require Java 8
This unblocks JUnit 5, at least, and shouldn't impose a burden on the consumers that I know about (all of which are well above the minimum SDK at which Java 8-targeted libs are supported).

benjamin-bader:
Closes #387 
Closes #388


benjamin-bader:
Breaking change: callbacks for oneway service methods now take `Unit`, not `Void`.

benjamin-bader:
More breaking changes:
- Protocol is now an interface
- Transport is now an interface

PR #395, opened 2020-08-24
Make antlr->kotlin dependency more elegant and efficient

benjamin-bader:


thrifty-schema/build.gradle
```diff
@@ -30,13 +30,6 @@ generateGrammarSource {
     arguments += ["-package", "com.microsoft.thrifty.schema.antlr"]
 }
 
-tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
```
benjamin-bader, line 33: this block is redundant with that in the root build.gradle file

Issue #406, opened 2020-12-02
Very large enums can generate Kotlin code that fails to compile


shashachu:
When you have thrift enums with lots of elements (say in the 3k-4k range), the Kotlin compiler will fail with an internal error:

```
Caused by: org.jetbrains.org.objectweb.asm.MethodTooLargeException: Method too large: com/foo/bar/BigEnum.<clinit> ()V
        at org.jetbrains.org.objectweb.asm.MethodWriter.computeMethodInfoSize(MethodWriter.java:2089)
        at org.jetbrains.org.objectweb.asm.ClassWriter.toByteArray(ClassWriter.java:458)
        at org.jetbrains.kotlin.codegen.ClassBuilderFactories$2.asBytes(ClassBuilderFactories.java:118)
        at org.jetbrains.kotlin.codegen.DelegatingClassBuilderFactory.asBytes(DelegatingClassBuilderFactory.kt:36)
        at org.jetbrains.kotlin.codegen.DelegatingClassBuilderFactory.asBytes(DelegatingClassBuilderFactory.kt:36)
        at org.jetbrains.kotlin.codegen.DelegatingClassBuilderFactory.asBytes(DelegatingClassBuilderFactory.kt:36)
        at org.jetbrains.kotlin.codegen.ClassFileFactory$ClassBuilderAndSourceFileList.asBytes(ClassFileFactory.java:348)
        at org.jetbrains.kotlin.codegen.ClassFileFactory$OutputClassFile.asByteArray(ClassFileFactory.java:313)
```

We found that removing the `Int` constructor param shrinks the `clinit` function sufficiently for the compile to succeed, i.e.

```
enum class BigEnum(val value: Int) {
    ONE(1),
    TWO(2),
    THREE(3),
    ...
}
```
will fail to compile, but
```
enum class BigEnum {
    ONE,
    TWO,
    THREE,
    ...
}
```
will succeed.

In order to make this work, we would also need to generate a `fun findByValue(value: BigEnum): Int?` function for serialization, and you'd lose the ability to easily convert to an `Int`.

Thoughts? Perhaps this could be a CLI option.

benjamin-bader:
Wow!  Can't say that I've ever butted up against method _body_ size limits before in this project.  I'm not surprised to hear that a 4k-member enum would do it.  Can you elaborate on how your enum could get so large?

[...]

PR #407, opened 2020-12-02
Add initial spike of big-enum support

benjamin-bader:
This is a rough draft of support for issue #406; at minimum, this also requires work to add a compiler flag to enable this mode (currently hardcoded).

@shashachu If you can, please give this a whirl and let me know if it works for you.  It's working for me, but I have a feeling that the integration tests we have are not sufficient to catch the kinds of bugs that would surface for something as big as your app.

### DIFF
--- a/thrifty-integration-tests/build.gradle
+++ b/thrifty-integration-tests/build.gradle
@@ -87,6 +87,7 @@ def kompileCoroutineTestThrift = tasks.register("kompileCoroutineTestThrift", Ja
             "--out=$projectDir/build/generated-src/thrifty/kotlin",
             "--lang=kotlin",
             "--kt-coroutine-clients",
+            "--experimental-kt-builderless-structs",
             "--kt-file-per-type",
             "$projectDir/CoroutineClientTest.thrift"
     ]
--- a/thrifty-integration-tests/build.gradle
+++ b/thrifty-integration-tests/build.gradle
@@ -87,7 +87,6 @@ def kompileCoroutineTestThrift = tasks.register("kompileCoroutineTestThrift", Ja
             "--out=$projectDir/build/generated-src/thrifty/kotlin",
             "--lang=kotlin",
             "--kt-coroutine-clients",
-            "--experimental-kt-builderless-structs",
             "--kt-file-per-type",
             "$projectDir/CoroutineClientTest.thrift"
     ]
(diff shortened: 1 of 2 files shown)


shashachu:
Wow, fast service! We'll give it a try internally and report back.

benjamin-bader:
Hi @shashachu - how did it go?  (or is it still ongoing?)

shashachu:
Hi, sorry I haven't had a chance to test it out yet.

shashachu:
Hey @benjamin-bader - so I've run into a couple issues using your snapshot on our codebase.

First, it doesn't seem to generate the proper imports for the `.value` extension function if the class it's referencing is in a different package than the current class. e.g.

```
package com.bar.foo

enum MyBigEnum {
    ...
}

val MyBigEnum.value: Int
    get() = ...
```

```
package com.foo.bar

// there will be a missing import for com.bar.foo.value

data class MyClass {
    ...
    override fun write(protocol: Protocol, struct: InterestReasonDetails): Unit {
      protocol.writeStructBegin("MyClass")
      if (struct.myBigEnum != null) {
        protocol.writeFieldBegin("myBigEnum", 1, TType.I32)
        protocol.writeI32(struct.myBigEnum.value)
        protocol.writeFieldEnd()
      }
     ...
    }
```

If I manually add the imports in, the build eventually fails with several `Duplicate JVM class name` errors, which appears to be this bug: https://youtrack.jetbrains.com/issue/KT-21186

I'm going to try implementing it as a regular member function instead of an extension function.

benjamin-bader:
Interesting!  Thanks for trying this out and reporting back.  I'm definitely curious about the how the member-function experiment works out.

shashachu:
@benjamin-bader Happy New Year! Finally got back to this work after the holidays, and it does appear to work when converting `value` into a regular function rather than an extension function. I honestly am unsure if we're simply kicking the can down the road, but everything builds for us.

I hacked it in essentially by inserting this code into `generateEnumClass`:

```
val value = FunSpec.builder("value")
  .returns(Int::class).apply {
      beginControlFlow("return when (this)")
      for (member in enumType.members) {
          val name = nameAllocator[member]
          addStatement("%L -> %L", name, member.value)
      }
      endControlFlow()
  }
  .build()

  ...

return typeBuilder
  .addType(companion)
  .addFunction(value)
  .build()
```

shashachu:
@benjamin-bader checking in again to see your thoughts about whether something like this could be merged into master. If not, we can certainly fork but we'd like to know sooner rather than later. Thanks!

benjamin-bader:
@shashachu thanks for your patience - life got pretty hectic this past month and I haven't checked notifications in a long time!

At this point, I'm happy to take a PR that helps you so long as it's behind a compiler flag.  Agreed  that this is probably just kicking the can down the road, but that's often a valid choice to make (and isn't my business in the first place).

benjamin-bader:
Implemented in #421

PR #414, opened 2020-12-18
Switch Kotlin structs to builderless by default

### LINKED ISSUE #409: Switch the default Kotlin struct style to builderless
Per the readme, Kotlin structs are still generated with `Builder` classes for backwards compatibility.  Since we're bumping from 2.x to 3.x, this is a fine opportunity to give people smaller code by default.

benjamin-bader:
Fixes #409

### DIFF
--- a/README.md
+++ b/README.md
@@ -412,7 +412,7 @@ and subject to change in future releases*.  It is a legacy from the time before 
 
 ## Kotlin Support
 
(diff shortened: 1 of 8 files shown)


benjamin-bader:
@jparise FYI - in the next major release, if you want to keep using Kotlin data classes that have `Builder`s, use the `--kt-struct-builders` compiler flag (or the corresponding Gradle plugin support, if you ever ended up getting that working).

PR #416, opened 2020-12-18
Location.base should always be an include root

### LINKED ISSUE #237: Location#path should have relative path to an include-path root
It seems that `Location#path` is always just `filename.thrift`; this is not very useful.  We'd intended for it to convey the relative position of the file, but from whose perspective?  Not that of the including file, because a file can be included more than once.

It should probably be relative to one of the paths in the include search-path (which includes the compiler's working directory, fyi).  Not sure how exactly to go about this, since we've no tests to verify this behavior and it has likely been broken for a very long time.

benjamin-bader:
As far as I can tell, Location has always been busted in practice - `base`, intended to be a "root" of some kind, has always been the full path to a file's parent directory, and "path" (intended to be a path relative to base) has just been a file name.

This makes Location little better than a string with line numbers in it - not all that useful.  In this PR, we fix what turns out to be the sole location (outside of tests) where Locations are created.  We do this by, when loading files from disk to parse, cycling through all of our include-roots, looking for the one with the shortest relative distance between the root and the file being loaded.

This isn't bulletproof since thrift files can include relative paths, and so potentially include files that aren't contained in any of our roots.  In that case, we'll keep the current behavior of calling the file's directory the location base.

This also isn't ideal - you could argue that, if a file under root A relatively includes a file under root B, and that included file wouldn't have otherwise been part of the compilation, that maybe A is the more useful attribution.  In the name of implementation simplicity, I disclaim that use case :)

@jparise I'd be grateful for your thoughts, if you've got the time and inclination.

Fixes #237

### DIFF
--- a/thrifty-schema/src/main/kotlin/com/microsoft/thrifty/schema/Loader.kt
+++ b/thrifty-schema/src/main/kotlin/com/microsoft/thrifty/schema/Loader.kt
@@ -191,8 +191,8 @@ class Loader {
                 return
             }
 
-            dir = file.parent
-            element = loadSingleFile(dir!!, file.fileName) ?: run {
+            dir = findClosestIncludeRoot(file) ?: file.parent!!
+            element = loadSingleFile(dir, dir.relativize(file)) ?: run {
                 val suffix = sourceElement?.let { "\n--> Included from ${it.location.filepath}" } ?: ""
                 throw FileNotFoundException("Failed to locate $path in $includePaths$suffix")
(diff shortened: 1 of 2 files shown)


jparise (approved):
I _think_ this looks correct.

I tried to consider cases where files with duplicate names exist in different include paths. Fortunately, this approach will still produce results consistent with the order of `includePaths`, so it shouldn't make an initially confusing situation any more confusing.

PR #419, opened 2021-01-04
Fix Code Generator emitting uncompilable code for exception named 'Exception'

luqasn:
by alias-importing the base type `kotlin.Exception`.
The root cause was that the builder tries to reference the outer type (the thrift `Exception`) but gets `kotlin.Exception` instead.

This introduces `kotest-assertions-compiler` for not just this, but all tests in `KotlinCodeGeneratorTest.kt` to check if the resulting code compiles.

Doing this still has the potential for a clash if you define a type `KotlinException` in thrift, but that is maybe more unlikely?

Looking forward to your feedback!


thrifty-kotlin-codegen/src/test/kotlin/com/microsoft/thrifty/kgen/util.kt
```diff
@@ -0,0 +1,15 @@
...
+import com.squareup.kotlinpoet.FileSpec
+import io.kotest.matchers.compilation.shouldCompile
+
+fun List<FileSpec>.shouldCompile() {
+    forEach {
+        it.shouldCompile()
+    }
+}
+
+fun FileSpec.shouldCompile() {
```
benjamin-bader, line 12: nit: we may as well put these in `KotlinCodeGeneratorTest.kt` since that's the only place they're used.  We can pull these out again if/when we need to.

benjamin-bader (changes requested):
Hi, thanks for the PR!  That's a rather unfortunate bug.  I'm not exactly sure how this aliased import is used, does it just hide `kotlin.Exception`?

It seems like a safer fix would be for generated code to use the fully-qualified name of the outer type.  We might be able to do that by explicitly importing `kotlin.Exception` - as I understand KotlinPoet, that _should_ be enough to surface the name clash and to force the use of the fully-qualified name.

Regardless of how we ultimately address the bug, I'll be grateful for getting the tests passing again.  Thanks again!

thrifty-kotlin-codegen/src/test/kotlin/com/microsoft/thrifty/kgen/util.kt
luqasn: sure, done

luqasn:
> Hi, thanks for the PR! That's a rather unfortunate bug. I'm not exactly sure how this aliased import is used, does it just hide `kotlin.Exception`?

this is what the generated code looks like with the aliasing:
```
package com.test

import com.microsoft.thrifty.Struct
import com.microsoft.thrifty.TType
import com.microsoft.thrifty.ThriftField
import com.microsoft.thrifty.kotlin.Adapter
import com.microsoft.thrifty.protocol.Protocol
import com.microsoft.thrifty.util.ProtocolUtil
import kotlin.String
import kotlin.jvm.JvmField
import kotlin.Exception as KotlinException

data class Exception(
  @JvmField
  @ThriftField(fieldId = 1)
  val message_: String?
) : KotlinException(), Struct {
  override fun write(protocol: Protocol) {
    ADAPTER.write(protocol, this)
  }

  private class ExceptionAdapter : Adapter<Exception> {
    override fun read(protocol: Protocol): Exception {
      var _local_message: String? = null
      protocol.readStructBegin()
      while (true) {
        val fieldMeta = protocol.readFieldBegin()
        if (fieldMeta.typeId == TType.STOP) {
          break
        }
        when (fieldMeta.fieldId.toInt()) {
          1 -> {
            if (fieldMeta.typeId == TType.STRING) {
              val message_ = protocol.readString()
              _local_message = message_
            } else {
              ProtocolUtil.skip(protocol, fieldMeta.typeId)
            }
          }
          else -> ProtocolUtil.skip(protocol, fieldMeta.typeId)
        }
        protocol.readFieldEnd()
      }
      protocol.readStructEnd()
      return Exception(message_ = _local_message)
    }

    override fun write(protocol: Protocol, struct: Exception) {
      protocol.writeStructBegin("Exception")
      if (struct.message_ != null) {
        protocol.writeFieldBegin("message", 1, TType.STRING)
        protocol.writeString(struct.message_)
        protocol.writeFieldEnd()
      }
      protocol.writeFieldStop()
      protocol.writeStructEnd()
    }
  }

  companion object {
    @JvmField
    val ADAPTER: Adapter<Exception> = ExceptionAdapter()
  }
}
```
 
> It seems like a safer fix would be for generated code to use the fully-qualified name of the outer type. We might be able to do that by explicitly importing `kotlin.Exception` - as I understand KotlinPoet, that _should_ be enough to surface the name clash and to force the use of the fully-qualified name.

Yes, that sounds safer, however I don't understand exactly what you mean by 'explicitly importing'. Can you provide some more guidance?

> Regardless of how we ultimately address the bug, I'll be grateful for getting the tests passing again. Thanks again!

Sure, I was a bit overeager at first and added the alias to all `FileSpec`s, thinking it would only apply when the type that is being aliases is used. That is not the case so tests that had assertions on the complete generated code would stumble over the new import (that was totally useless). I fixed that.

Maybe related (but maybe not): https://github.com/square/kotlinpoet/issues/938

benjamin-bader:
> Maybe related (but maybe not): square/kotlinpoet#938

Good find!  This does indeed seem related.  The recent comment by Zac seems to indicate that the solution I was proposing may not actually work.  Let me do a bit of exploration - I still feel like there's a safer option than import aliases here.

luqasn:
Definitely an upstream issue, I found out the following funny repro case:
```
import com.squareup.kotlinpoet.ClassName
import com.squareup.kotlinpoet.FileSpec
import com.squareup.kotlinpoet.FunSpec
import com.squareup.kotlinpoet.TypeSpec

val namespace = "test"

val kotlinExceptionName = ClassName("kotlin", "Exception")
val customExceptionName = ClassName(namespace, "Exception")
val customException = TypeSpec
        .classBuilder("Exception")
        .superclass(kotlinExceptionName)
        .addFunction(
                FunSpec
                        .builder("test")
                        .addParameter("e", customExceptionName)
                        .build()
        )
        .build()


FileSpec.builder(namespace, "a.kt")
        .addType(customException)
        .build()
        .toString()

val funn = FunSpec.builder("test")
        .addParameter("e", customExceptionName)
        .addParameter("f", kotlinExceptionName)
        .build()

FileSpec.builder(namespace, "b.kt")
        .addType(customException)
        .addFunction(funn)
        .build()
        .toString()
```

this outputs
```
package test

import kotlin.Exception

class Exception : Exception() {
  fun test(e: Exception) {
  }
}
```

and

```
package test

class Exception : kotlin.Exception() {
  fun test(e: Exception) {
  }
}

fun test(e: Exception, f: kotlin.Exception) {
}
```

So as soon as I use both colliding types in the same function, the anti-collision works as expected. 🤷 

I will report this upstream.

luqasn:
I validated that the upstream fix actually fixes this issue - unfortunately no release yet. This can be closed - in case you want to keep the `shouldCompile` assertions this also introduced, I can open another PR with only that.

benjamin-bader:
Sorry, life got hectic and I didn't check notifications for quite some time now.

I see you made the upstream fix - well done!  I had started one and was making it way more complicated than your solution ended up being.  Will be very glad to pull in the new kotlinpoet version.  If you're up for another PR for the `shouldCompile` assertions I'd be glad to have those, too.

PR #421, opened 2021-02-04
Adding support for alternative Kotlin enum generation

shashachu:
When generating very very large enums, the generated code will fail to compile due to JVM code-size limits (see #406).

This PR (based on #407) adds a compiler flag to use an alternative method to generate Kotlin enum classes, which works around this limitation by removing the `value` member and instead adds a `value()` function to map an enum to its Int value. 

Since this flag also generates a `value` property which calls through to the `value()` function, this change can be done transparently to Kotlin callers.

### DIFF
--- a/thrifty-compiler/src/main/kotlin/com/microsoft/thrifty/compiler/ThriftyCompiler.kt
+++ b/thrifty-compiler/src/main/kotlin/com/microsoft/thrifty/compiler/ThriftyCompiler.kt
@@ -59,7 +59,7 @@ import java.util.ArrayList
  * [--kt-file-per-type]
  * [--kt-struct-builders]
  * [--kt-jvm-static]
- * [--kt-big-enum]
+ * [--kt-big-enums]
  * [--parcelable]
  * [--use-android-annotations]
  * [--nullability-annotation-type=[none|android-support|androidx]]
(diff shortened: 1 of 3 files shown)


benjamin-bader (approved):
LGTM!

benjamin-bader:
Thanks @shashachu for taking it home!
I'm evaluating release plans for v3.0.0; it will hopefully be out soon-ish and will contain this fix.

benjamin-bader:
Hmm, merged too soon.  This seems to me like it won't work with serialization - writing out enums will use the `getValue()` function (via the inferred `value` property), and not this new `value()` function.  This will _appear_ to work but, if your enums aren't numbered the way Java assumes, then you'll probably end up with incorrect values on the wire.

The fix will be for generated adapters, in their `write` method, to call `value()` instead of `value`.

shashachu:
@benjamin-bader sorry for not making that clearer. If you think it's *too* clever, I'm ok with removing the `value` property. To be frank, it was mostly laziness with not wanting to change every single existing call-site in our code that references `value`. 😅

shashachu:
Also - @benjamin-bader - do you have a more specific timeline for 3.0.0 than 'soon-ish'? Totally understandable if not, as 3.0.0 looks like a pretty major refactor; just trying to decide if we should wait for 3.0.0 or temporarily use a local jar. Appreciate all your help with this!

benjamin-bader:
I definitely do not have a fixed timeline, and this is certainly a major refactor.  The criteria are more to the effect of,
a) Is the code stable for prod usage?
b) Has more than one major user reported the same?

I intend to shore up some of the in-progress work and push 3.0.0-RC01 within a week or so.  We'll see how it goes from there :)

benjamin-bader:
No, I don't think it's _too_ clever; it's surprising, yes, but not the sort of thing that is likely to cause headaches IMO.  I really only see this causing issues in super weird cases like reflectively getting an enum's value, or maybe using this stuff from Java sources somehow.

Also, I figure that if your project is big enough to need this, then you probably have a good number of specialists on hand and can readily solve whatever comes up.

PR #444, opened 2021-02-24
Fix builder class generating name clash if union contains field named "error"

luqasn:
Unfortunately, there seems to be an upstream bug in addition to the missing fully qualified reference to kotlin.error in the code generator.

Or do you have an idea why my fix would not work?

### DIFF
--- a/thrifty-kotlin-codegen/src/test/kotlin/com/microsoft/thrifty/kgen/KotlinCodeGeneratorTest.kt
+++ b/thrifty-kotlin-codegen/src/test/kotlin/com/microsoft/thrifty/kgen/KotlinCodeGeneratorTest.kt
@@ -191,6 +191,21 @@ class KotlinCodeGeneratorTest {
         xception.propertySpecs.single().name shouldBe "message_"
     }
 
+
+    @Test
+    fun `union with error as field name should compile with builder`() {
+        val thrift = """
+            namespace kt com.test
+
+            union Fail { 1: string error }
+        """.trimIndent()
+
+        val specs = generate(thrift) {
+            withDataClassBuilders()
+        }
+        specs.shouldCompile()
+    }
+
     @Test
     fun services() {
         val thrift = """
--- a/thrifty-kotlin-codegen/src/test/kotlin/com/microsoft/thrifty/kgen/KotlinCodeGeneratorTest.kt
+++ b/thrifty-kotlin-codegen/src/test/kotlin/com/microsoft/thrifty/kgen/KotlinCodeGeneratorTest.kt
@@ -84,6 +84,7 @@ class KotlinCodeGeneratorTest {
 
         val files = KotlinCodeGenerator(FieldNamingPolicy.JAVA).generate(schema)
 
+        files.shouldCompile()
         files.forEach { println("$it") }
     }
 
@@ -127,6 +128,7 @@ class KotlinCodeGeneratorTest {
         val gen = KotlinCodeGenerator().filePerType()
         val specs = gen.generate(schema)
 
+        specs.shouldCompile()
         specs.single().name shouldBe "Constants" // ".kt" suffix is appended when the file is written out
     }
 
@@ -139,6 +141,7 @@ class KotlinCodeGeneratorTest {
         """.trimIndent()
 
         val specs = generate(thrift)
+        specs.shouldCompile()
 
         val struct = specs.single().members.single() as TypeSpec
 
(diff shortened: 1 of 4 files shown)


benjamin-bader:
kotlinpoet hasn't yet had a release containing your fix; are you testing with a local build to know that there's an issue?

It seems like there's an active _situation_ around Dokka and the move from jcenter to Maven Central, so CI isn't even running the tests you've added.  I'll see if I can clear that up.

benjamin-bader:
OK, now CI shows your test-case failing.

As far as why your fix isn't working, my guess is that it's because `kotlin.error` is a function, not a type, and thus wouldn't be affected by the regular CodeWriter import rules.  Just a guess though.  Top-level functions are a rather late addition to kotlinpoet.

luqasn:
Yeah, I think you are right. This seems to be a new, albeit similar upstream bug, using master does not fix it.

luqasn:
> I think a better, if less-satisfying, fix will be to update KotlinCodeGenerator to reserve the name `error`. We'd do that in the function that generates `NameAllocator` instances, around line 155 or so. This will result in the field being named something like `error_` which kind of sucks, but will at least compile :)

Better as in quicker? Because to me, this is a clear KotlinPoet bug again. If I fully-qualify the reference I expect it to stay fully qualified if there is potential for a clash.

benjamin-bader:
Better as in quicker, yes.  I share your expectation.

luqasn:
There is no time pressure for the bug, so I would rather fix it upstream (unless extremely complicated or something). Will have a look what I can find.

luqasn:
Not the prettiest solution: https://github.com/square/kotlinpoet/pull/1040

luqasn:
Marked this as ready for merge because thrifty's side is done, now we are waiting for a new KotlinPoet version including my fix. An alternative to waiting until we can merge would be to merge the required code (the fully qualified method) and test in `@Disabled` state right now so we don't keep this open.

thrifty-kotlin-codegen/src/test/kotlin/com/microsoft/thrifty/kgen/KotlinCodeGeneratorTest.kt
```diff
@@ -357,7 +357,7 @@ class KotlinCodeGeneratorTest {
 
         file.single().toString() should contain("""
             |public interface Svc {
-            |  public suspend fun doSomething(foo: Int): Int
+            |  public abstract suspend fun doSomething(foo: Int): Int
```
benjamin-bader, line 360: I'm so confused.  Why does a generated interface need to include the `abstract` keyword?  Isn't it implicit in the fact the the function belongs to an interface and has no body?

luqasn: Totally agree with you. KotlinPoet is a bit weird here. Although the redundant modifier is only cosmetic, I am curious and checking out the relevant code :)

benjamin-bader: I see; I guess it may be another "library-mode" change.  That sure junked up our generated code.  Guess it's a net win though.

luqasn: The weird thing is not that it now emits the keyword when you explicitly specify it (that sounds very much correct).
The weird thing is that you can not NOT explicitly make a function in an interface abstract (which is why thrifty also needs to add an abstract marker).
The notion of implicit abstractness of a interface function does not exist.

benjamin-bader:
@luqasn looks like kotlinpoet 1.8.0 just shipped, including your recent bugfixes!

luqasn:
Moved the updating to https://github.com/microsoft/thrifty/pull/446, will rebase after that's merged

luqasn:
Ready to merge now. Thankfully, the KotlinPoet folks even took the very ugly "interface methods emit explicit abstract modifier" back :)

benjamin-bader (approved):

benjamin-bader:
Thanks for seeing this through - I really appreciate how thorough you've been!

luqasn:
> Thanks for seeing this through - I really appreciate how thorough you've been!

You are more than welcome! Thanks for being so responsive and supportive!

PR #447, opened 2021-04-10
Add Server support

luqasn:
Excited to have this in a shareable state and very interested in hearing your feedback :)

I created separate commits for refactorings that were required for the new code that then gets added "append-only" in one larger commit to make it easier to gauge the impact (rather non-impact) on the existing code.

One thing that is not perfect yet is that the synthetic arg and result structs are generated on the same level as the normal ones, opening the possibility for a name clash. Ideally, I would want them nested inside the `ServerCall` class, but that required changes to the existing data class/union generators, so I skipped it for now.


luqasn:
One other thing that is a bit suboptimal is that the default error handler does not log anything when an error occurs. I didn't add that because up until now there was no dependency on any logging framework in thrifty-runtime and I was hesitant to add one. But it might be worth it, what do you think?

luqasn:
Side note: I am pretty confident that apart from the default requiredness thing (which is very minor), Thrifty is 1 to 1 compatible/equivalent with the thrift implementation.
You cited the lack of isSet* properties as a source of difference and non-conformity making Thrifty unfit for the server side, but I am pretty sure that's only the case if you were to use non-boxed primitives (as is the case in default Java thrift), as the isSet* implementation for reference types is just a null check on the field in default Java thrift.
But with Kotlin, the primitives are boxed anyways (at least
If optional, not sure about non-optional), so a null check is always equivalent to an isSet call.

luqasn:
Friendly nudge ;)

thrifty-integration-tests/src/test/kotlin/com/microsoft/thrifty/integration/conformance/server/TestServer.kt
```diff
@@ -0,0 +1,131 @@
...
+import java.io.OutputStream
+
+import java.io.IOException
+
+import com.sun.net.httpserver.HttpExchange
+import kotlinx.coroutines.runBlocking
+import okio.Buffer
+
+
+class TestServer @JvmOverloads constructor(private val protocol: ServerProtocol = ServerProtocol.BINARY) : Extension, BeforeEachCallback, AfterEachCallback {
```
benjamin-bader, line 55: Curious at the use of `@JvmOverloads`.  Do you expect this to be used from Java?

thrifty-integration-tests/src/test/kotlin/com/microsoft/thrifty/integration/conformance/server/ThriftTestHandler.kt
```diff
@@ -0,0 +1,179 @@
...
+
+
+        /*
+     *   { 1 => { 2 => argument,
+     *            3 => argument,
+     *          },
+     *     2 => { 6 => <empty Insanity struct>, },
+     *   }
+     */
+        val result: MutableMap<Long, Map<com.microsoft.thrifty.integration.kgen.coro.Numberz, com.microsoft.thrifty.integration.kgen.coro.Insanity>> = LinkedHashMap()
```
benjamin-bader, line 115: If possible can we import these types?

thrifty-kotlin-codegen/src/main/kotlin/com/microsoft/thrifty/kgen/KotlinCodeGenerator.kt
```diff
@@ -62,6 +62,7 @@ import com.microsoft.thrifty.service.AsyncClientBase
 import com.microsoft.thrifty.service.MethodCall
 import com.microsoft.thrifty.service.ServiceMethodCallback
 import com.microsoft.thrifty.service.TMessageType
+import com.microsoft.thrifty.service.server.*
```
benjamin-bader, line 65: nit: we generally try to avoid wildcard imports.

thrifty-kotlin-codegen/src/main/kotlin/com/microsoft/thrifty/kgen/KotlinCodeGenerator.kt
```diff
@@ -311,6 +317,21 @@ class KotlinCodeGenerator(
             }
         }
 
+        if (generateServer) {
+            check(coroutineServiceClients) { "Server feature currently requires coroutineServiceClients to be enabled" }
```
benjamin-bader, line 321: Have you considered having the generate do _either_ clients _or_ servers?

thrifty-kotlin-codegen/src/main/kotlin/com/microsoft/thrifty/kgen/KotlinCodeGenerator.kt
```diff
@@ -1974,6 +2002,174 @@ class KotlinCodeGenerator(
...
+            spec.addCode {
+                addStatement( "%S -> %T()",
+                        method.name,
+                        ClassName(serviceType.kotlinNamespace, serverTypeName, callTypeName(method)))
+            }
+        }
+
+        spec.addCode {
+            beginControlFlow("else -> ")
+            addStatement("""throw %T("%L")""", IllegalArgumentException::class, "Unknown method \${msg.name}")
```
benjamin-bader, line 2062: Be careful with using `Foo::class` as type arguments.  Kotlinpoet has an iffy track record with understanding, for example `kotlin.Exception` - historically it will often end up emitting references to the underlying JVM type, which is _ok_ I guess for this use-case but forecloses on MPP servers too.

tl;dr let's add an entry to the `ClassNames` object up around line 110 of this file and use that instead.

benjamin-bader: I see a few other places in the PR (and probably in my code, too, let's be honest) where `::class` literals are being used this way.  So long as they aren't typealiases or expect/actual types, it's probably OK, but we should make sure.

thrifty-runtime/src/commonMain/kotlin/com/microsoft/thrifty/ThriftException.kt
```diff
@@ -30,8 +30,8 @@ import kotlin.jvm.JvmStatic
  */
 class ThriftException(
         @JvmField
-        val kind: Kind?,
-        message: String?) : RuntimeException(message) {
+        val kind: Kind,
+        message: String?) : RuntimeException(message), Struct {
```
benjamin-bader, line 34: nice, I should have thought of this long ago :)

benjamin-bader (commented):
Hey, I'm very sorry for the delay.  This honestly just fell off my radar after a very hectic month at work.

This is a great piece of work!  I quibble about some minor things, but not enough to block on.  I'd like to get this landed and refine as needed.

thrifty-integration-tests/src/test/kotlin/com/microsoft/thrifty/integration/conformance/server/TestServer.kt
luqasn: This file was based on the existing `TestServer.java` that I converted to Kotlin using IntelliJ, so it was so helpful as to automatically add the annotation :)
I've removed it now.

thrifty-integration-tests/src/test/kotlin/com/microsoft/thrifty/integration/conformance/server/ThriftTestHandler.kt
luqasn: yep!

thrifty-kotlin-codegen/src/main/kotlin/com/microsoft/thrifty/kgen/KotlinCodeGenerator.kt
luqasn: fixed

thrifty-runtime/src/commonMain/kotlin/com/microsoft/thrifty/ThriftException.kt
luqasn: Wasn't really necessary because you were never writing it :)

thrifty-kotlin-codegen/src/main/kotlin/com/microsoft/thrifty/kgen/KotlinCodeGenerator.kt
luqasn: Yeah, good point. I wanted to add as little stuff as possible and thus thought it would be a good idea to reuse the client interface as the service interface (server side), because the signatures are exactly going to match.
But because it did not make much sense to me to implement the callback based variant for the server side, I chose to force the coroutine client flag so that the interface would be generated with coroutine suspend keywords and that's what the server generator will generated against.
It is kinda hacky though and you are totally right, I would suspect you either generate the client or the server, but rarely both at the same time.
For this reason and also because of the weird interaction between client and server implementations, it would make more sense to generate a separate interface just for the server and then have that either blocking or suspending controlled via its own flag (in case you want to build a non-blocking thrift server e.g. using ktor), but not generate a callback-based implementation.

That means I need to copy the method that generates the client interface and adjust it for server generation. Also not great, but better than the current state I guess.

thrifty-kotlin-codegen/src/main/kotlin/com/microsoft/thrifty/kgen/KotlinCodeGenerator.kt
luqasn: Honestly I have almost no experience with Kotlin Multiplatform, so I tried to do what you said to the places I found :)

luqasn:
Sorry for the long delay, was struggling to find the motivation to sit down after work to get this done :)
Tried to take care of all the feedback, the `coroutineServiceClient` thingy will need to wait for another evening, though it should not require huge changes.

thrifty-kotlin-codegen/src/main/kotlin/com/microsoft/thrifty/kgen/KotlinCodeGenerator.kt
luqasn: This should be fixed now.

luqasn:
Server generation now always generates its own service interface in a `server` namespace that is always coroutine-based.
First, I tried for allowing both "plain" service handlers and coroutine ones, but that was a pain in the back because I had to duplicate a lot of code because suspend functions have different signatures.
So in the end I opted to just always generate coroutine based servers, because it allows for more freedom for the user, e.g. to use `async` constructs, `either` magic from Arrow or whatever.
If you choose to not use any coroutine specific stuff in your handler, that should be totally fine and not impact performance much, all you have is an extraneous parameter on our methods for the coroutine continuation (or maybe the compiler even optimizes that away if you never suspend).

luqasn:
I have addressed your feedback (thanks a lot!) and added the missing license headers.

benjamin-bader:
Nice, thank you!  This generally looks good.  I'm going to hold off on merging this, however, until after we finalize a 3.0.0 release.  That should be fairly soon, but I'm still working out some problems with it.

benjamin-bader:
A whole month later, I've _finally_ released 3.0.0.  Thanks for your extraordinary patience.  Let's get this landed!

luqasn:
I have pulled your changes in via a rebase fixing any conflicts and only after I was done did I realize that that's not the most transparent way for you to track what has changed. Was totally in my work-workflow, sorry. I guess you would have probably preferred a merge commit?

README.md
```diff
@@ -346,13 +347,30 @@ Builders are unnecessary, and are not included by default.  For compatibility wi
...
 `none` (the default), `android-support`, and `androidx`.  Specifying `android-support` will cause generated code to use `@Nullable` and `@NonNull` from
 the `android.support.annotation` package.  Similarly, specifying `androidx` will use analogous annotations from `androidx.annotation`. 
 
 ## Thanks
+`--experimental-kt-generate-server` enabled code generation for the server portion of a thrift service. You can
+use this to implement a thrift server with the same benefits as the kotlin client: no runtime surprises thanks to
+structs being always valid by having nullability guarantees and unions represented as sealed classes.
+See 
+
+### Thanks
```
benjamin-bader, line 373: This readme block looks misplaced.  Is this an artifact of the rebase?

luqasn: yes, seems I somehow messed that up. Fixed now.

luqasn:
Bump :)

benjamin-bader:
oh geez, were you waiting for me?  I thought I was waiting for _you_ 😧 
let me take a look again

benjamin-bader (approved):
:shipit:

Issue #370, opened 2020-07-02
Support for THttpClient Transport


cyber-gh:
Are there any plans to support the THttpClient from the original thrift implementation?

benjamin-bader:
No current plans, but that's only because we haven't had that need.  I'm not opposed to it, but don't personally have the time or inclination to do the work myself.  I'd be happy to help guide a PR.

cyber-gh:
I will give it a try, but this may take some time, as I have to document on this in topic

clayreimann:
@benjamin-bader I'm also interested in seeing this happen (especially given the server-side thrift stuff that looks to be landing soon). Can you point me to the places that I should start looking to build the new transport?

benjamin-bader:
Most transports live in https://github.com/microsoft/thrifty/tree/master/thrifty-runtime/src/commonMain/kotlin/com/microsoft/thrifty/transport.

For convenience you might start putting your code here, but I see a few issues to be resolved prior to merging:
- this is "common" code and most HTTP libraries are platform-specific.  Maybe OkHttp is multiplatform now?
- a main goal of this project is to be as lightweight as possible in terms of method references.  It may be desirable to put this transport into its own module.

luqasn:
I think this can be close now as it was added in https://github.com/microsoft/thrifty/pull/449 :)

benjamin-bader:
Done!

PR #449, opened 2021-06-18
Add http transport

luqasn:
addresses https://github.com/microsoft/thrifty/issues/370

Implementation based on the original thrift http transport with a few changes:
* apache http client variant removed
  * it was weird to have multiple different implementations in the same class
  * I did not want to introduce an additional dependency
  * okio is already a dependency, so maybe we want to base the transport on that?
* tried to make the code a bit more kotlin idiomatic and safer

Even though there are no tests yet I wanted to collect some early feedback, especially on okio vs. HTTPUrlConnection. This needs to live in the jvm variant due to the availability of HTTPUrlConnection.


thrifty-runtime/src/jvmMain/kotlin/com/microsoft/thrifty/transport/THttpTransport.kt
```diff
@@ -0,0 +1,172 @@
...
+ * implementations to switch between in the same class.
+ *
+ * Uses HttpURLConnection internally
+ *
+ * Also note that under high load, the HttpURLConnection implementation
+ * may exhaust the open file descriptor limit.
+ *
+ * @see [THRIFT-970](https://issues.apache.org/jira/browse/THRIFT-970)
+ */
+open class THttpTransport(url: String) : Transport {
```
benjamin-bader, line 45: Also, a substantive comment - can we please make an `expect class` in `src/commonMain/kotlin/com/microsoft/thrifty/transport` for this?  Then make this one the `actual`?

Generally on JVM I favor using OkHttp over HttpURLConnection - it's much easier to use correctly and scalably.  Is there a reason, aside from the initial implementation using it, to favor HttpURLConnection?

I suppose the fact that it's built in to the JDK is a point in its favor, along with not wanting to bring in extra dependencies.  This is part of why I suggested making a separate module for this guy.

thrifty-runtime/src/jvmMain/kotlin/com/microsoft/thrifty/transport/TTransportException.kt
```diff
@@ -0,0 +1,6 @@
+package com.microsoft.thrifty.transport
+
+class TTransportException : Throwable {
```
benjamin-bader, line 3: I tend to think this exception doesn't buy us anything, and would prefer not to include it.  We could use the perfectly-good `ProtocolException` in its stead.

benjamin-bader (changes requested):
This is a great start - thanks for picking this up!  There are a few things I think we could change to make this fit seamlessly in with the rest of the project; noted inline.  Also, I'd really like to see some sort of testing here.  OkHttp's `MockWebServer` could be very handy for this, assuming it's still maintained and available (been a few years since I used it).

thrifty-runtime/src/jvmMain/kotlin/com/microsoft/thrifty/transport/TTransportException.kt
luqasn: `ProtocolException` unfortunately does not have the possibility to pass a `cause` (doesn't have that constructor implemented), and I really don't like wrapping exceptions in a way that loses information.
But I do agree with you that this does not provide much value, so I will go for just not catching and rethrowing these wrapped.

thrifty-runtime/src/jvmMain/kotlin/com/microsoft/thrifty/transport/THttpTransport.kt
luqasn: What benefit do we get from making it an expect/actual class? Why can't there just be different types of Transports implemented for different variants? This is a genuine question, I have no experience with Kotlin Multiplatform.

luqasn: Yes, main reason is OOB-ness. I don't want to add something to the runtime that drags in a bunch of transitive dependencies. For that I would prefer the separate module. But I see nothing wrong with providing a basic HttpTransport with maybe not the best performance that needs no dependencies and is this simple as part of the "core" library.

luqasn:
> This is a great start - thanks for picking this up! There are a few things I think we could change to make this fit seamlessly in with the rest of the project; noted inline. Also, I'd really like to see some sort of testing here. OkHttp's `MockWebServer` could be very handy for this, assuming it's still maintained and available (been a few years since I used it).

Yes, I am definitely planning to write tests still!

thrifty-runtime/src/jvmMain/kotlin/com/microsoft/thrifty/transport/THttpTransport.kt
benjamin-bader: Fair point re: "OOB-ness".  My concern definitely is around the default easy path.  Nothing's easier, dependency-wise, than what's built in.  On the other hand, this is an Android library first and foremost, and on Android OkHttp is basically the standard anymore.  I'm OK with leaving it out for now.

As far as expect/actual, this is the mechanism that lets us have interfaces in common code but have platform-specific implementations.  In the eventual iOS release, we'll definitely want to use an `HttpTransport`.  Having an `expect class HttpTransport` in common code is a prerequisite.

thrifty-runtime/src/jvmMain/kotlin/com/microsoft/thrifty/transport/TTransportException.kt
benjamin-bader: OK, SGTM.

thrifty-runtime/src/jvmMain/kotlin/com/microsoft/thrifty/transport/THttpTransport.kt
luqasn: I was trying to understand the benefit of the `expect` over a plain interface for this usecase. We can already have different implementations of the http transport, they would all just implement `Transport`.

So if I understand correctly, `expect` just tells you "whatever platform this is ported to also needs to implement this thing", which to me is mainly useful for the "common" library to be able to call platform specific code without having the user pass an implementation to some constructor explicitly (there is always going to be exactly one implementation, no more, no less).
To me, plain old interface still sounds totally fitting to me for this use case. Consider the case that we add an `OkHttpClient` in a separate library as well. Would that then work with the `expect`? Or would this now be a "plain" implementation of `Transport`? If yes, why treat both implementations differently?

Please excuse all the text, I am mainly trying to wrap my head around Kotlin Multiplatform.

benjamin-bader: You're right that we _could_ just have platform-specific types that implement `Transport`.  The thing I think lacks there is that we won't have a single `HttpTransport`.  That seems like a miss to me.

benjamin-bader: I never addressed you question about the benefit(s) of `expect class` over an interface.  The difference is that an `expect class` is _concrete_ - one can instantiate one as-is, without needing to consider what other type might implement the interface.

juequinterore:
Hi.

I would like to ask, is there any plan to continue with this work?

luqasn:
Sorry for the late reply. I picked this up again and have integration tests ready, but there are still issues with the plumbing (extension initialization order).
Will try to finish this ASAP.

juequinterore:
Thanks for your work @luqasn!

I have been using this code since couple of months ago and it has been
working without any issue.

luqasn:
Long time no see, @benjamin-bader :)

I finally finished the integration tests I promised, could you have a look? Thanks!

benjamin-bader:
Hey, glad to see you @luqasn!  The test code is pretty substantial - thanks for digging in to it.  I don't know why the tests didn't run earlier, given that you've made PRs here before, but I just had to re-approve running the test suite.  It failed the license-header check, which I fixed here.

I'm inclined to take this large PR as-is.  Thanks very much for all of your hard work on it, and for coming back to finish it!

I might come back and refactor `HttpTransport` into an expect/actual type, so that we can use this in iOS (we have some rudimentary iOS support now), but that's somewhere off in the future.

benjamin-bader (approved):

benjamin-bader:
Also fair warning for anyone following this PR - Microsoft has changed the way they handle publishing permissions to Maven Central, but started the tooling migration before Github support was done.  This leaves Thrifty temporarily unable to make releases.  I hear that in another month or so, this should be fixed, but in the meantime please be patient.

luqasn:
Thanks for letting us know, I think after those two years one month more is not gonna bother anyone ;)

Anyone in a pinch could also use jitpack.io to get a "prerelease" version of thrifty-runtime. Brilliant service, though I can't vouch for their security. Be mindful of that.

PR #459, opened 2021-08-10
Added huge enum generation

shashachu:
For very very large enums, even the `--kt-big-enum` workaround is not enough to avoid the Java code size limits, so we've added a `--kt-huge-enum` option which generates a hierarchy of enum-like classes that are syntactically equivalent to Kotlin the `enum class`.

### DIFF
--- a/thrifty-compiler/src/main/kotlin/com/microsoft/thrifty/compiler/ThriftyCompiler.kt
+++ b/thrifty-compiler/src/main/kotlin/com/microsoft/thrifty/compiler/ThriftyCompiler.kt
@@ -126,7 +126,8 @@ import kotlin.system.exitProcess
  * `--kt-huge-enums` is optional.  When specified, generated enums will use a different
  * representation. Instead of `enum class`, a hierarchy of enum-like classes will be generated
  * that is syntactically equivalent to access from Kotlin  This should
- * be avoided unless you know you need it.  Implies `--lang=kotlin`.
+ * be avoided unless you know you need it.  Implies `--lang=kotlin` and mutually-exclusive with
+ * `--kt-big-enums`.
  *
  * `--parcelable` is optional.  When provided, generated types will contain a
  * `Parcelable` implementation.  Kotlin types will use the `@Parcelize` extension.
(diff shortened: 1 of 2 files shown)


shashachu:
oops didn't mean to put this PR on the microsoft fork.

benjamin-bader:
I understand this wasn't meant to go here, but... bravo, that's really clever!

jparise:
> I understand this wasn't meant to go here, but... bravo, that's really clever!

If you think there's a path to including it in mainline Thrifty, we can definitely do that. 🙂 

Otherwise, the plan is to maintain the branch on our public fork.

shashachu:
@benjamin-bader haha thanks! This strategy was mostly figured out by other engineers here; I just adapted it for Kotlin.

FWIW I did come up with a different strategy that is, as far as I know, 'infinitely' scalable and syntactically equivalent but only from Kotlin. It takes advantage of the fact that in Kotlin, companion objects can have superclasses. Since we do still have some Java in our project, we unfortunately couldn't use it smoothly. In Java you would need to add a `.Companion` call which isn't the end of the world but is ugly:

```
For this example enum:
enum Color {
  RED = 1,
  ORANGE = 2,
  YELLOW = 3,
  GREEN = 4,
  BLUE = 5,
  PURPLE = 6
}

We would generate:

open class DO_NOT_USE_ME_Parent_Color_0 protected constructor() {
    @JvmField
    val RED = Color("RED", 1)
    @JvmField
    val ORANGE = Color("ORANGE", 2)

    open fun findByValue(`value`: Int): Color? = when (`value`) {
        1 -> RED
        2 -> ORANGE
        else -> null
    }
}

open class DO_NOT_USE_ME_Parent_Color_1 protected constructor() : DO_NOT_USE_ME_Parent_Color_0() {
    @JvmField
    val YELLOW = Color("YELLOW", 3)
    @JvmField
    val GREEN = Color("GREEN", 4)

    override fun findByValue(`value`: Int): Color? = when (`value`) {
        3 -> YELLOW
        4 -> GREEN
        else -> super.findByValue(value)
    }
}

open class DO_NOT_USE_ME_Parent_Color_2 protected constructor() : DO_NOT_USE_ME_Parent_Color_1() {
    @JvmField
    val BLUE = Color("BLUE", 5)
    @JvmField
    val PURPLE = Color("PURPLE", 6)

    override fun findByValue(`value`: Int): Color? = when (`value`) {
        5 -> BLUE
        6 -> PURPLE
        else -> super.findByValue(value)
    }
}

class Color(val name: String, val value: Int) {
    companion object : DO_NOT_USE_ME_Parent_Color_2()
    override fun toString(): String = name
}
```

PR #463, opened 2022-01-20
Upgrade to Kotlin  1.6.10

benjamin-bader:
This one's a doozy.  For whatever reason, Kotlin 1.6 induces a Gradle issue where pure-Java modules (`thrifty-kotlin-codegen`, `thrifty-compiler`, etc) can no longer properly reference the mutliplatform `thrifty-runtime` project.  Something to do with ambiguous variants - oh by the way, did you know Grade modules have variants? - with names like `apiElements` vs `jvmApiElements`.

Documentation on this area of Gradle is sparse.  I _think_ the right thing to do is to advertise slightly different attributes for the JVM-specific variants, but honestly am not sure.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -21,4 +21,4 @@ POM_DEVELOPER_URL=https://www.bendb.com
 # Dokka is a greedy little pig
 org.gradle.jvmargs=-XX:MaxMetaspaceSize=512m
 
-KOTLIN_VERSION=1.5.32
+KOTLIN_VERSION=1.6.10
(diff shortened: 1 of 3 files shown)


benjamin-bader:
Okay!  Looks like, per https://github.com/google/guava/pull/3683#issuecomment-786729042, Android projects don't care specifically about the attributes we're setting on `thrifty-runtime`'s `jvmApiElements` and `jvmRuntimeElements` variants.  If I'm reading this correctly, they shouldn't interfere with variant selection in Android and thus are safe to ship.

PR #464, opened 2022-02-01
Update ThriftyCompiler.kt

Jimexist:
the default lang type is Kotlin not Java


benjamin-bader:
lol good catch - thanks for the contribution!

PR #486, opened 2022-05-19
Add generated comments to Kotlin files (default true)

### LINKED ISSUE #485: Question/feature request: Add header/comments to each file
I poked through the code and docs and didn't see anything - is there a way to add a common header to every generated file? We've had a couple folks not realize the files are auto-generated and try to hand-edit them. There used to be the `@Generated` annotation but that got removed when the project moved to Kotlin MPP, but being able to add a Javadoc header would be even better.

shashachu:
Will respect the existing `--omit-file-comments` flag

Fixes #485


thrifty-kotlin-codegen/src/test/kotlin/com/microsoft/thrifty/kgen/KotlinCodeGeneratorTest.kt
```diff
@@ -1346,8 +1351,25 @@ class KotlinCodeGeneratorTest {
...
+        file.shouldCompile()
+
+        val lines = file.toString().split("\n")
+        lines[0] shouldBe "// Automatically generated by the Thrifty compiler; do not edit!"
+        lines[1] shouldContain "\\/\\/ Generated on: [0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}\\.\\S+".toRegex()
+    }
+
     private fun generate(thrift: String, config: (KotlinCodeGenerator.() -> KotlinCodeGenerator)? = null): List<FileSpec> {
-        val configOrDefault = config ?: { this }
+        val configOrDefault = config ?: { emitFileComment(false) }
```
shashachu, line 1372: I can also default this to true but it's a bit annoying to handle the changing timestamp.

thrifty-kotlin-codegen/src/main/kotlin/com/microsoft/thrifty/kgen/KotlinCodeGenerator.kt
```diff
@@ -149,6 +151,7 @@ class KotlinCodeGenerator(
     private var emitJvmName: Boolean = false
     private var emitJvmStatic: Boolean = false
     private var emitBigEnums: Boolean = false
+    private var emitFileComment: Boolean = true
```
benjamin-bader, line 154: Personally I'd prefer this to default to false, _if only_ to minimize the disruption to others.  AFAIK most people using thrifty are still manually checking in the generated code.

thrifty-kotlin-codegen/src/main/kotlin/com/microsoft/thrifty/kgen/KotlinCodeGenerator.kt
```diff
@@ -2606,5 +2617,12 @@ class KotlinCodeGenerator(
     private fun String.capitalize(): String {
         return this.replaceFirstChar { if (it.isLowerCase()) it.titlecase(Locale.US) else it.toString() }
     }
+
+    companion object {
+        private const val FILE_COMMENT =
```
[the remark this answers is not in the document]
benjamin-bader: I'd have imagined that after working in this codebase, you'd have seen enough copypasta from me not to worry about it 😆 

(This is absolutely fine by me, thank you for asking)

benjamin-bader:
Sorry for the delay; crazy family visit == no time to think about anything!  I appreciate the effort.  This basically LGTM but I'd like to make the comment an opt-in feature - that is, it should default to `false`.

thrifty-kotlin-codegen/src/main/kotlin/com/microsoft/thrifty/kgen/KotlinCodeGenerator.kt
shashachu: @benjamin-bader I don't have a strong opinion either way, but it does default to true for Java sources, which means the default will change depending on whether you generate Java or Kotlin. I will defer to you.

shashachu:
@benjamin-bader How dare you prioritize family over your volunteer OSS work! :) Anyway, I responded to your comment re the default value. I can update the docs if we end up having different behavior for Java vs Kotlin.

thrifty-kotlin-codegen/src/main/kotlin/com/microsoft/thrifty/kgen/KotlinCodeGenerator.kt
benjamin-bader: You make a good point.  I'd forgotten entirely about the Java code-generator, despite directly referring to it.  Shows where our focus is, I suppose.  Let's keep it on by default, then.

benjamin-bader:
> How dare you prioritize family over your volunteer OSS work! :)

It's one weakness that I indulge, I admit.

Thanks for your contributions, as always!

Issue #523, opened 2023-01-22
Support comments for struct properties


a8t3r:
```thrift
struct Query {
  /**
  * what are you looking for
  */
  1: required string text,
  2: optional i64 resultsNewerThan
}
```

should be:
```kotlin
public data class Query(
    /**
    * what are you looking for
    */
    @JvmField
    @ThriftField(fieldId = 1, required = true)
    public val text: String
 ...
)
```

I think this feature should be disabled by default for backward compatibility

benjamin-bader:
Hi @a8t3r!  Thanks for the reminder - I'd completely forgotten that we dont emit property-level kdoc.  We did this for Java in 1.0 but demurred on Kotlin.  It's been a long time since I looked at that part of the code, but IIRC I had a notion that we'd want to use `@property` tags in the data class' top-level doc comment.  It _might_ be easy to do, or it might be fiddly.  Certainly we could just slap the documentation on the property itself and call it a day.  (if so, I _think_ we'd need to do that [here](https://github.com/microsoft/thrifty/blob/master/thrifty-kotlin-codegen/src/main/kotlin/com/microsoft/thrifty/kgen/KotlinCodeGenerator.kt#L551) but am not positive - would need to check kotlinpoet's behavior, it might need to go on the `ParameterSpec` declared above also/instead)

Agreed that we could gate that behind a compiler flag/plugin DSL function.

Honestly I won't get to this anytime soon, but I'd be thrilled to take a PR.

PR #524, opened 2023-01-23
ISSUE-523 Support comments for struct properties

a8t3r:
https://github.com/microsoft/thrifty/issues/523

1) I discovered java generator produces documentation for class and fields by default, so I think it would be better to save same semantic for kotlin generator - so there are no additional flags

2) Fixed issue with ignored `structBuilders`


a8t3r:
@microsoft-github-policy-service agree

benjamin-bader (approved):
LGTM

benjamin-bader:
Thanks, looks great - and thanks for finding + fixing the builder bug in the gradle plugin, too.

PR #528, opened 2023-02-13
Refactor code-coverage report into a plugin

benjamin-bader:
While adding MPP-specific source-sets to the coverage report, it became clear that the coverage report was getting quite complex, and that Groovy was limiting my confidence in correctness.  Refactoring this task into a build-src plugin increases that confidence, just a bit.

### DIFF
--- a/build-src/build.gradle
+++ b/build-src/build.gradle
@@ -32,6 +32,13 @@ gradlePlugin {
             description = "Maven publication logic"
             implementationClass = "com.microsoft.thrifty.ThriftyPublishPlugin"
         }
+
+        coverage {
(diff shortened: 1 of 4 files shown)


benjamin-bader:
On further inspection, this change does _not_ get thrifty-runtime code into coverage reports.  After a bunch of experimentation, I can only conclude that kotlin-mpp just doesn't play well with jacoco, which is also not flexible enough in Gradle to fix easily.

We don't get nearly enough value from these reports to justify the effort in fixing them, so I'm closing this PR without merging.  I'm kind of inclined to drop them entirely.

PR #538, opened 2023-07-26
Pass name style policy to kotlin code generator

### LINKED ISSUE #537: Gradle plugin ignores FieldNameStyle specification for Kotlin code gen
Hi Thrifty devs!

I was trying to implement thrifty in my android project using the [Gradle plugin](https://github.com/microsoft/thrifty/tree/master/thrifty-gradle-plugin) so that we can benefit from the language features of Kotlin (null-safety etc). 

I'm currently using regular apache-thrift to generate java thrifts, and we specify that our field names should be generated using `camelCase`. So to make migration to Thrifty easier I'm trying to use the `nameStyle` field. However this doesn't appear to be working as expected as I'm still seeing the fields using the `DEFAULT` styling (as-is in thrift source).

I did some digging and it looks like perhaps the plugin is not passing the `nameStyle` input to the `KotlinCodeGenerator`, so it just defaults to the `DEFAULT` style.

```groovy
thrifty {
    kotlin {
        nameStyle 'java'
    }
}
```

rhappe:
This fix should make it so that the kotlin code generation respects the `nameStyle` option when using the gradle plugin.

Fixes #537


benjamin-bader:
Thanks very much for the contribution!

PR #548, opened 2024-01-17
Overrides for strictRead and strictWrite in BinaryProtocol

### LINKED ISSUE #547: BinaryProtocol hard-codes strict read/write properties
`BinaryProtocol` hard-codes both the `strictRead` and `strictWrite` properties to be `false` with no valid way to override them. Not that this matters much since microsoft/thrifty can set whatever defaults that are seen to be fit, but it is also different from the [defaults set by apache-thrift:](https://github.com/apache/thrift/blob/3dcc0cdeb993c363f14bc6514cff32f9ac5b6ac7/lib/java/src/main/java/org/apache/thrift/protocol/TBinaryProtocol.java#L93).

Suggestion is to keep the defaults as `false` for both if it was intentional, but at least provide a way to override them in the  `BinaryProtocol` constructor.

rhappe:
- Adding overrides for the `strictRead` and `strictWrite` in the `BinaryProtocol` constructor. Before these were defaulted to `false` and immutable.
- Tests for `beginReadMessage` and `beginWriteMessage`, which utilize the `strictRead` and `strictWrite` values.

Fixes #547


benjamin-bader (approved):
LGTM - thanks very much!

rhappe:
@microsoft-github-policy-service agree

Issue #540, opened 2023-09-10
Fork!


benjamin-bader:
Following an internal policy change at Microsoft, this project has no maintainers able to make releases.

I'm going to continue work in my own personal fork of the repository at github.com/benjamin-bader/thrifty.  Will need to repackage it to facilitate publication on Maven Central

francescotescari:
Unfortunate :( we use still use thrift in our project and wanted to explore KMP (ios, android, web) using this lib. I see that there are more recent commits on this repo compared to the fork, is the fork just for releases? If I want to contribute, which repo should I work on?

benjamin-bader:
Good question.  The fork is in limbo at the moment - it's possible that publishing rights will be restored.  I'm not clear on the details yet.  Please pick this one if you'd like to contribute, and if the situation doesn't resolve then I can port your work over.

PR #551, opened 2024-05-29
Fix that const definitions referencing imported consts cannot be loaded

### LINKED ISSUE #550: Const definitions referencing imported consts cannot be loaded
I have the following two thrift files:
___
`path/to/errors.thrift`
```thrift

const string ERROR_NOT_FOUND = "ErrorNotFound"
...
```
___
`service.thrift`
```thrift
include "path/to/errors.thrift"

const string NOT_FOUND = errors.ERROR_NOT_FOUND
...
```
___
While with Apache Thrift Compiler, I can generate code out of these two thrift files without issues, Thrifty failed to load `service.thrift` with the following error message:
```
E: Unrecognised const identifier: errors.ERROR_NOT_FOUND (at path/to/service.thrift)
```

Right now my workaround is to define `NOT_FOUDN` in `service.thrift` with literal `"ErrorNotFound"` instead of the imported reference `errors.ERROR_NOT_FOUND`.

CoolTomatos:
Fixes microsoft/thrifty#550

### DIFF
--- a/thrifty-schema/src/main/kotlin/com/microsoft/thrifty/schema/Linker.kt
+++ b/thrifty-schema/src/main/kotlin/com/microsoft/thrifty/schema/Linker.kt
@@ -443,10 +443,9 @@ internal class Linker(
             if (ix != -1) {
                 val includeName = symbol.substring(0, ix)
                 val qualifiedName = symbol.substring(ix + 1)
-                val expectedPath = "$includeName.thrift"
                 constant = program.includes
                         .asSequence()
-                        .filter { p -> File(p.location.path).name == expectedPath }
+                        .filter { p -> p.location.programName == includeName }
                         .mapNotNull { p -> p.constantMap[qualifiedName] }
                         .firstOrNull()
             }
--- a/thrifty-schema/src/main/kotlin/com/microsoft/thrifty/schema/Linker.kt
+++ b/thrifty-schema/src/main/kotlin/com/microsoft/thrifty/schema/Linker.kt
@@ -446,7 +446,7 @@ internal class Linker(
                 val expectedPath = "$includeName.thrift"
                 constant = program.includes
                         .asSequence()
-                        .filter { p -> p.location.path == expectedPath } // TODO: Should this be ==, or endsWith?
+                        .filter { p -> File(p.location.path).name == expectedPath }
                         .mapNotNull { p -> p.constantMap[qualifiedName] }
                         .firstOrNull()
             }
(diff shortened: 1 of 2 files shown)


CoolTomatos:
@microsoft-github-policy-service agree company="Uber"

thrifty-schema/src/main/kotlin/com/microsoft/thrifty/schema/Linker.kt
```diff
@@ -446,7 +446,7 @@ internal class Linker(
                 val expectedPath = "$includeName.thrift"
                 constant = program.includes
                         .asSequence()
-                        .filter { p -> p.location.path == expectedPath } // TODO: Should this be ==, or endsWith?
+                        .filter { p -> File(p.location.path).name == expectedPath }
```
benjamin-bader, line 449: If you feel up to it, I think it would make sense to add `val name` to `Location` itself, rather than implement it here in this lambda.

benjamin-bader (changes requested):

thrifty-schema/src/main/kotlin/com/microsoft/thrifty/schema/Linker.kt
CoolTomatos: Hi @benjamin-bader, I noticed that there is an existing API on Location for checking this. So I changed to use that instead. Please take another look.

thrifty-schema/src/main/kotlin/com/microsoft/thrifty/schema/Linker.kt
```diff
@@ -443,10 +443,9 @@ internal class Linker(
             if (ix != -1) {
                 val includeName = symbol.substring(0, ix)
                 val qualifiedName = symbol.substring(ix + 1)
-                val expectedPath = "$includeName.thrift"
                 constant = program.includes
                         .asSequence()
-                        .filter { p -> p.location.path == expectedPath } // TODO: Should this be ==, or endsWith?
+                        .filter { p -> p.location.programName == includeName }
```
benjamin-bader, line 448: Haha nice, I totally forgot about `programName`.  Good find!

benjamin-bader (approved):

CoolTomatos:
Hi @benjamin-bader, I've noticed that we haven't cut a release for a few years now. 
I have also noticed #540 
May I ask what the plan is going forward?

benjamin-bader:
Great question!  I don't have an answer for you.  I'm still unable to make releases.  My internal contact claimed that they were working something out, but it's been six months and silence since then.  I'd held off on really pushing the fork forward, but don't feel very hopeful about this repo.

$dayjob keeps me 110% busy, so very little time remains for OSS.  I'd encourage you, if you need this today, to use your own fork for the time being.  When time and life permit, I will probably get back to my fork.

CoolTomatos:
Thanks a lot for the reply @benjamin-bader.

